Issue #27, opened 2018-08-13
Update to Network Testing


mdhorn:
<a href="https://github.com/mdhorn"><img src="https://avatars.githubusercontent.com/mdhorn" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mdhorn](https://github.com/mdhorn)**
_Sunday Jul 08, 2018 at 01:09 GMT_

----

- Start test when entering Testing screen automatically
- Test --> ReTest
- If Network test passes, then remove the 'Cancel' button and default to selecting the 'Done' button.
- If any network changes are made to the configuration, unmark Testing as completed (remove the '+')
  - Change the selected network device
  - Change of Proxy URL
  - Change of the Swupd Mirror URL

dorileo:
@mdhorn I think this issue is completed now with the #36 PR merged.

mdhorn:
Closed by PR #36 and PR #42

PR #36, opened 2018-08-31
Network test popup

mdhorn:
Fixes Issue: #27 

- Switch to using a model PopUp window for Network testing.
- Added URL checking to the Proxy setting

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -30,6 +30,12 @@ import (
 	"github.com/clearlinux/clr-installer/utils"
 )
 
+var (
+	// NetworkPassing is used to track if the latest network configuration
+	// is passing; changes in proxy, etc.
+	NetworkPassing bool
+)
+
 func sortMountPoint(bds []*storage.BlockDevice) []*storage.BlockDevice {
 	sort.Slice(bds[:], func(i, j int) bool {
 		return filepath.HasPrefix(bds[j].MountPoint, bds[i].MountPoint)
@@ -245,15 +251,17 @@ func ConfigureNetwork(model *model.SystemInstall) error {
 	prg, err := configureNetwork(model)
 	if err != nil {
 		prg.Success()
+		NetworkPassing = false
 		return err
 	}
+
+	NetworkPassing = true
+
 	return nil
 }
 
 func configureNetwork(model *model.SystemInstall) (progress.Progress, error) {
-	if model.HTTPSProxy != "" {
-		cmd.SetHTTPSProxy(model.HTTPSProxy)
-	}
+	cmd.SetHTTPSProxy(model.HTTPSProxy)
 
 	if len(model.NetworkInterfaces) > 0 {
 		prg := progress.NewLoop("Applying network settings")
--- a/tui/common.go
+++ b/tui/common.go
@@ -52,6 +52,11 @@ type Page interface {
 }
 
 const (
+	// WindowWidth is our desired terminal width
+	WindowWidth = 80
+	// WindowHeight is our desired terminal width
+	WindowHeight = 24
+
 	// MenuButtonPrefixUncompleted is the standard, uncompleted prefix for a menu button
 	MenuButtonPrefixUncompleted = "[ ]"
 	// MenuButtonPrefixCompletedByConfig is the completed by config prefix for a menu button
@@ -180,6 +185,15 @@ const (
 	// ActionNone indicates no action has been performed
 	ActionNone
 )
+
+// Is the page id a PopUp page
+func isPopUpPage(id int) bool {
+	if id == TuiPageNetworkValidate {
+		return true
+	}
+
+	return false
+}
 
 // GetConfigDefinition is a stub implementation
 // the real implementation must check with the model and return:
@@ -313,10 +327,10 @@ func (page *BasePage) setup(tui *Tui, id int, btns int, returnID int) {
 func (page *BasePage) newWindow() {
 	sw, sh := clui.ScreenSize()
 
-	x := (sw - 80) / 2
-	y := (sh - 24) / 2
+	x := (sw - WindowWidth) / 2
+	y := (sh - WindowHeight) / 2
 
-	page.window = clui.AddWindow(x, y, 80, 24, " [Clear Linux Installer ("+model.Version+")] ")
+	page.window = clui.AddWindow(x, y, WindowWidth, WindowHeight, " [Clear Linux Installer ("+model.Version+")] ")
 	page.window.SetTitleButtons(0)
 	page.window.SetSizable(false)
 	page.window.SetMovable(false)
--- a/tui/menu.go
+++ b/tui/menu.go
@@ -6,8 +6,11 @@ package tui
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/VladimirMarkelov/clui"
+
+	"github.com/clearlinux/clr-installer/controller"
 )
 
 // MenuPage is the Page implementation for the main menu page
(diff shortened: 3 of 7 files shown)


tui/common.go
```diff
@@ -52,6 +52,11 @@ type Page interface {
 }
 
 const (
+	// TermWidth is our desired terminal width
+	TermWidth = 80
```
dorileo: this is not actually the terminal size, but the window size.

controller/controller.go
```diff
@@ -30,6 +30,12 @@ import (
 	"github.com/clearlinux/clr-installer/utils"
 )
 
+var (
+	// NetworkPassing is used to track if the latest network configuration
+	// is passing; changes in proxy, etc.
+	NetworkPassing bool
```
dorileo: we could use the model to store this flag, we don't even need to export it anyway.

controller/controller.go
```diff
@@ -245,15 +251,17 @@ func ConfigureNetwork(model *model.SystemInstall) error {
...
+	NetworkPassing = true
+
 	return nil
 }
 
 func configureNetwork(model *model.SystemInstall) (progress.Progress, error) {
-	if model.HTTPSProxy != "" {
-		cmd.SetHTTPSProxy(model.HTTPSProxy)
-	}
+	cmd.SetHTTPSProxy(model.HTTPSProxy)
```
dorileo: what happens if a proxy is not set?

tui/menu.go
```diff
@@ -101,7 +104,24 @@ func newMenuPage(tui *Tui) (Page, error) {
...
+		if controller.NetworkPassing {
+			// Network already validated, just start the install
+			page.GotoPage(TuiPageInstall)
+		} else {
+			// Network needs to be validated before the install
+			if dialog, err := CreateNetworkTestDialogBox(page.tui.model); err == nil {
+				if dialog.RunNetworkTest() {
+					// Automatically close if it worked
+					clui.RefreshScreen()
+					time.Sleep(time.Second)
```
dorileo: Do you really need this sleep here?

tui/network_dialog_check.go
```diff
@@ -0,0 +1,216 @@
...
+	resultLabel   *clui.Label
+	dismissButton *SimpleButton
+}
+
+// Success is part of the progress.Client implementation and represents the
+// successful progress completion of a task by setting
+// the progress bar to "full"
+func (dialog *NetworkTestDialog) Success() {
+	dialog.progressBar.SetValue(dialog.progressMax)
+	clui.RefreshScreen()
```
dorileo: You're refreshing the screen twice (here and in the menu callback), is that intentional?

tui/network_dialog_check.go
```diff
@@ -0,0 +1,216 @@
...
+	bg := dialog.progressBar.BackColor()
+	flashTime := 100 * time.Millisecond
+	//dialog.progressBar.SetValue(0)  // leave the bar where it fails?
+	for i := 1; i <= 5; i++ {
+		dialog.progressBar.SetBackColor(term.ColorRed)
+		clui.RefreshScreen()
+		time.Sleep(flashTime)
+		dialog.progressBar.SetBackColor(bg)
+		clui.RefreshScreen()
+		time.Sleep(flashTime)
```
dorileo: sleeping and refreshing twice?

tui/network_dialog_check.go
```diff
@@ -0,0 +1,216 @@
...
+	} else {
+		dialog.progressBar.Step()
+	}
+	clui.RefreshScreen()
+}
+
+// Desc is part of the progress.Client implementation and sets the progress bar label
+func (dialog *NetworkTestDialog) Desc(desc string) {
+	dialog.resultLabel.SetTitle(desc)
+	clui.RefreshScreen()
```
dorileo: Also refreshing? are you running this dialog in a different goroutine?

tui/network_dialog_check.go
```diff
@@ -0,0 +1,216 @@
...
+// LoopWaitDuration is part of the progress.Client implementation and returns the time duration
+// each step should wait until calling Step again
+func (dialog *NetworkTestDialog) LoopWaitDuration() time.Duration {
+	return 1 * time.Second
+}
+
+// OnClose sets the callback that is called when the
+// dialog is closed
+func (dialog *NetworkTestDialog) OnClose(fn func()) {
+	clui.WindowManager().BeginUpdate()
```
dorileo: You shouldn't need this bend and end update pairs, specially here.

tui/network_dialog_check.go
```diff
@@ -0,0 +1,216 @@
...
+	defer clui.WindowManager().EndUpdate()
+	dialog.onClose = fn
+}
+
+// Close closes the dialog window and executes a callback if registered
+func (dialog *NetworkTestDialog) Close() {
+	clui.WindowManager().DestroyWindow(dialog.DialogBox)
```
dorileo: ditto

tui/network_dialog_check.go
```diff
@@ -0,0 +1,216 @@
...
+	if posY < hBuff {
+		posY = hBuff
+	}
+	posY = y + posY
+
+	dialog.DialogBox = clui.AddWindow(posX, posY, dWidth, dHeight, title)
+	dialog.DialogBox.SetTitleButtons(0)
+	dialog.DialogBox.SetMovable(false)
+	dialog.DialogBox.SetSizable(false)
+	clui.WindowManager().BeginUpdate()
```
dorileo: again?

tui/network_dialog_check.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil, fmt.Errorf("Failed to create Network Test Dialog: %v", err)
+	}
+
+	if modelSI == nil {
+		return nil, fmt.Errorf("Missing model for Network Test Dialog")
+	}
+	dialog.modelSI = modelSI
+
+	dialog.dismissButton.OnClick(func(ev clui.Event) {
+		go func() {
```
dorileo: you don't need this to be in a separate go routine

dorileo:
@mdhorn all those RefreshScreen() calls might have been inherited from my original code, I think it's a good opportunity to review'em.

tui/common.go
mdhorn: Renaming to WindowWidth and WindowHeight

controller/controller.go
mdhorn: @dorileo But is "the network working" really part of the model of the install?  To me, it was a status of something during execution. But if you feel strongly I can relocate it.

controller/controller.go
mdhorn: If it is empty, it is set to empty and works.
I've tested this code both with a value, empty, new value and back to empty -- works.
In fact, there was a corner case where the old value was left because even though the model was updated to be empty, the cmd still had the old proxy value.

tui/menu.go
mdhorn: It was only there to give the user a "second" to see the Success status before the dialog automatically goes away. I added it because it bothered me that the popup went away too quickly. If you have a strong feeling about this, we can remove it.

tui/network_dialog_check.go
mdhorn: Who doesn't like a Fresh a screen? :)  I can remove one of these and retest.

tui/network_dialog_check.go
mdhorn: This is the "Flashing" between Red and the default color to make the failure obvious to the user. We have the same code in Failure code for the Progress bar during the installation code.

tui/network_dialog_check.go
mdhorn: This refresh after setting the Title was imitated from the Dialog window code in clui.

tui/network_dialog_check.go
mdhorn: This Begin and End was imitated from the Dialog window code in clui.

tui/network_dialog_check.go
mdhorn: ditto

tui/network_dialog_check.go
mdhorn: This is how it is code in the PopUp dialog code in clui.

tui/network_dialog_check.go
mdhorn: Good catch. Removed.

tui/network_dialog_check.go
mdhorn: @dorileo We are also doing a refresh after SetValue in the tui/installer code and after SetTitle; this seems to be the BKM pattern, no?

dorileo:
@mdhorn I've just tested the patches and it breaks when resizing terminal. If you hit "test network" and resize the terminal then the TUI window is not properly placed/moved, I also saw some window corruption.

dorileo:
if we fail the testing after changing the proxy settings we shouldn't revert the user's change.

dorileo:
if we fail the testing after changing the network settings we shouldn't revert the user's change.

mdhorn:
@dorileo After testing proxy settings, if testing fails, we do reset the proxy to the previous value.

mdhorn:
@dorileo Can you please re-review now that I have pushed changes.
Are there any concerns which still need to be addressed or discussed?

PR #38, opened 2018-09-05
Vendoring update

mdhorn:
- Refactor makefile commands to be more go-like
- Updated Dependencies to latest versions, including tip of master for clui


dorileo:
@mdhorn I suggest we wait a little while wit this re vendoring, Vladmir replied my latest pull request and we might have a new set of changes coming today or tomorrow.

mdhorn:
@dorileo I was hoping to get merged the refresh fix which is now in their master and these Makefile changes which make new updates very simple. This would also enable me to merge my other pending PR with the "official" refresh fix.
I'd be very happy to do a second PR Friday or Monday with the new scrolling frame fixes too.

PR #40, opened 2018-09-10
controller: don't reconfigure network

dorileo:
Now we have the network check dialog and we don't need/want to re
test the network in the install process.


mdhorn:
@dorileo Replacing with PR #42

PR #41, opened 2018-09-10
Makefile updates

mdhorn:
- Ensure files are copied with same permissions
- New target for cleaning testing cache
- New target for running tests as root


mdhorn:
@dorileo This fixes the problem with running the checker has root and then having to sudo make clean  -- I do this testing a lot before submitting a PR

PR #42, opened 2018-09-10
Optimize Network Testing

mdhorn:
- Do not rerun network test if network known to be passing
- Run Network PopUp testing if any Network Interface is modified

Replaces the open PR #40

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -94,8 +94,11 @@ func Install(rootDir string, model *model.SystemInstall) error {
 		return err
 	}
 
-	if err = ConfigureNetwork(model); err != nil {
-		return err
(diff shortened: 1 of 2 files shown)


controller/controller.go
```diff
@@ -94,8 +94,10 @@ func Install(rootDir string, model *model.SystemInstall) error {
 		return err
 	}
 
-	if err = ConfigureNetwork(model); err != nil {
-		return err
+	if !NetworkPassing {
```
dorileo: can you add a comment telling we have the mass installer use case?

mdhorn: @dorileo yes, adding now.

mdhorn:
@dorileo Does the new comment suffice?

PR #43, opened 2018-09-11
Testing the Travis-CI

mdhorn:
Want to know what UID the code is checkout out to so
we can add a user with the same UID for testing the
code as a non-root user.


dorileo:
seems it didn't work though.

mdhorn:
@dorileo yes, not ready for review. and lower priority. I'll add you as reviewer when ready.

mdhorn:
Abandoning as the latest travis works well and we now have local coverage checking.

Issue #53, opened 2018-09-12
Allow Escape-key to "cancel" current menu


mdhorn:
Consider adding support for using the Escape-key to back up to the previous menu - same as Cancel button.

PR #61, opened 2018-09-19
Enable Escape key to cancel to previous page

mdhorn:
Fixes Issue: #53

- Trap on the Escape-key on the common page type; reacts like the default 'Cancel' key to return to the "previous" page as defined by page setup.

### DIFF
--- a/tui/common.go
+++ b/tui/common.go
@@ -327,6 +327,19 @@ func (page *BasePage) setup(tui *Tui, id int, btns int, returnID int) {
 		"Use [Tab] or the arrow keys [up and down] to navigate", Fixed)
 
 	page.window.SetVisible(false)
+
+	// Escape-key cancel's this screen and returns
+	// same as the default 'Cancel' button
+	page.window.OnKeyDown(func(ev clui.Event, data interface{}) bool {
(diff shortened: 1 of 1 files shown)


tui/common.go
```diff
@@ -327,6 +327,19 @@ func (page *BasePage) setup(tui *Tui, id int, btns int, returnID int) {
 		"Use [Tab] or the arrow keys [up and down] to navigate", Fixed)
 
 	page.window.SetVisible(false)
+
+	// Escape-key cancel's this screen and returns
+	// same as the default 'Cancel' button
+	page.window.OnKeyDown(func(ev clui.Event, data interface{}) bool {
```
dorileo: looks good and works fine, I have only one question, what would be the expected behavior if you're typing in a text field and press escape?

mdhorn: The same as pressing the Cancel button.
If there are text fields where we think this is a concern/risk, we can trap and remove the Escape from the input stream.
But we will no be using vi/emacs like editing in our text fields. :)

dorileo: I would rather make it Ctrl+Escape though

Issue #18, opened 2018-08-13
Customize kernel cmdline


mdhorn:
<a href="https://github.com/asifhg"><img src="https://avatars.githubusercontent.com/asifhg" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [asifhg](https://github.com/asifhg)**
_Monday May 07, 2018 at 18:38 GMT_

----

Allow the user to completely edit the default kernel cmdline, instead of only appending to it.

mdhorn:
<a href="https://github.com/dorileo"><img src="https://avatars.githubusercontent.com/dorileo" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [dorileo](https://github.com/dorileo)**
_Monday May 07, 2018 at 19:41 GMT_

----

We don't currently have a kernel cmdline configuration yet - I'm sure ister has it but the new installer hasn't received this configuration yet.

I'll make sure to allow editing instead of appending whenever we have it.

mdhorn:
<a href="https://github.com/dorileo"><img src="https://avatars.githubusercontent.com/dorileo" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [dorileo](https://github.com/dorileo)**
_Tuesday Jun 19, 2018 at 18:06 GMT_

----

I've implemented in the PR #150 the feature to append kernel arguments, we could evolve from there to edit - I haven't saw the detail to implement such change yet.

mdhorn:
Making inquiries if there are specific use cases in mind that can not be accommodated by just adding to the existing command line.

bryteise:
@mdhorn So editing the commandline is a bit of a pain still but you can remove things from it once I make a new release with https://github.com/clearlinux/clr-boot-manager/pull/124 in it.

mdhorn:
@bryteise Very helpful to know. Thank you!

mdhorn:
@asifhg The ability to add to or remove the kernel command line during the installation was implemented in version 1.0.0.

PR #63, opened 2018-09-20
kernel arguments: add "remove kernel arguments" support

dorileo:
We are relying on clr-boot-manager's cmdline-removal feature which
prevents a kernel argument to be added to the boot manager kernel
entry.

Fixes Issue: #18

- Add remove kernel command line argument support

### DIFF
--- a/tui/kernel_cmdline.go
+++ b/tui/kernel_cmdline.go
@@ -5,36 +5,76 @@
 package tui
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/VladimirMarkelov/clui"
 )
 
 // KernelCMDLine is the Page implementation for the kernel cmd line configuration page
 type KernelCMDLine struct {
 	BasePage
-	kernelCMDLineEdit *clui.EditField
+	addKernelArgEdit *clui.EditField
+	remKernelArgEdit *clui.EditField
 }
+
+const (
+	kernelArgsHelp = `Note: The boot manager tool will first include the "Add Extra Arguments"
+      items then the "Remove Arguments" items are removed from the final
+      - where the final list contains the kernel bundle's configured
+      arguments and the ones configured by the user.`
+)
 
 // GetConfiguredValue Returns the string representation of currently value set
 func (pp *KernelCMDLine) GetConfiguredValue() string {
-	cmdLine := pp.getModel().KernelCMDLine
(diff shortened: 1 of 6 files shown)


network/network_test.go
```diff
@@ -50,6 +50,12 @@ func TestGatewayLoading(t *testing.T) {
 	}
 }
 
+func TestVerifyConnectivity(t *testing.T) {
```
mdhorn: Why are we change the network test coverage in this pull request?

tui/kernel_cmdline.go
```diff
@@ -5,36 +5,66 @@
...
+			values = append(values, fmt.Sprintf("add: %s", addKernelArgs))
+		}
+
+		if remKernelArgs != "" {
+			values = append(values, fmt.Sprintf("remove: %s", remKernelArgs))
+		}
 
-	if cmdLine == "" {
-		return "No kernel command line append set"
+		result = strings.Join(values, " - ")
```
mdhorn: The dash/hyphen here seems confusing to me; it could be a part of the add command.
Could we use something that isn't valid for the kernel command line like `|` or ...

tui/kernel_cmdline.go
```diff
@@ -5,36 +5,66 @@
...
+	result := ""
+
+	if pp.getModel().KernelArguments != nil {
+		values := []string{}
+
+		addKernelArgs := strings.Join(pp.getModel().KernelArguments.Add, " ")
+		remKernelArgs := strings.Join(pp.getModel().KernelArguments.Remove, " ")
+
+		if addKernelArgs != "" {
+			values = append(values, fmt.Sprintf("add: %s", addKernelArgs))
```
mdhorn: Maybe make the labels "ADD:" and "REMOVE:"? Or at least "Add:" and "Remove:"

network/network_test.go
dorileo: I'll drop this.

tui/kernel_cmdline.go
dorileo: done

tui/kernel_cmdline.go
dorileo: done

dorileo:
@mdhorn added the priority note.

PR #64, opened 2018-09-20
Change Done to Confirm

mdhorn:
- Another UX review suggestion implemented


dorileo:
I wouldn't change the SetDone() API, it reflects more the menu or step status, marking a step as done. The buttons is a different thing alongside the btn flag.

Issue #48, opened 2018-09-11
"iconify" menu settings line


dorileo:
**Describe the bug**
Currently it's clipping if the settings line is too long, we should make it more friendly and "iconify" that line.

**To Reproduce**
N/A

**Expected behavior**
N/A

**Screenshots**
N/A

**Environment (please complete the following information):**
N/A

**Additional context**
N/A

mdhorn:
Closed by #65

PR #65, opened 2018-09-20
tui: iconify menu's configuration line

dorileo:
Fixes Issue: #48 

- iconify menu's configuration line

### DIFF
--- a/tui/menu_button.go
+++ b/tui/menu_button.go
@@ -6,6 +6,7 @@ package tui
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/VladimirMarkelov/clui"
 	xs "github.com/huandu/xstrings"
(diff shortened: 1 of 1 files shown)


tui/menu_button.go
```diff
@@ -142,11 +142,16 @@ func (mb *MenuButton) Draw() {
 	clui.PushAttributes()
 
 	itemValue := fmt.Sprintf("%c %s", statusSymbol[mb.status], mb.itemValue)
-	shift, itemText := clui.AlignColorizedText(itemValue, w, mb.Align())
+
+	if len(itemValue) >= w {
+		itemValue = fmt.Sprintf("%s...", clui.CutText(itemValue, w-7))
```
mdhorn: Why the magic number 7?
Testing this by enabling all current bundles results in only one black space after the "..."; given the magic number 7 and then putting in three '.'s plus on space -- there seem to be 4 characters still missing?
Can we make this a calculation?

dorileo: done

PR #67, opened 2018-09-21
Errors validation error

dorileo:
- Fixes a small issue with the check-coverage.sh script
- Introduces an specialized error for validation
- Nicely display model validation errors
- Add validation for kernel configuration (kernel config is mandatory)

To test, run mass installer with an invalid configuration (i.e no system language set).

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -20,6 +20,17 @@ type TraceableError struct {
 	What  string
 }
 
+// ValidationError is a type of error used to report model or any general condition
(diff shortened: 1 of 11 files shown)


errors/errors.go
```diff
@@ -20,6 +20,12 @@ type TraceableError struct {
 	What  string
 }
 
+// ValidationError is a type of error
```
mdhorn: Please add a comment explaining why we have a ValidationError type, and when it should be used.

dorileo: done

Issue #68, opened 2018-09-21
Telemetry should be started with telemctl, not systemctl


mdhorn:
**Describe the bug**
The latest version of the telemetry server split up the services, so the old name being used by 'systemctl' is no longer valid in the latest Clear Linux release.

**Expected**
Use the 'telemctl' for all state changes to telemetry services.

PR #70, opened 2018-09-21
Use Telemetry control, not system

mdhorn:
Fixes Issue: #68

- Switch to using the telemetry package control program


mdhorn:
@dorileo I verified the tests are passing on both an older Clear Linux OS system with the old telemetry services and on a latest Clear Linux OS.

PR #73, opened 2018-09-25
local-travis: add a travis "emulation"

dorileo:
This patch introduces the local-travis, an small util program locally
emulate the remote travis execution. The program will read the travis
yaml file and execute the tasks.

The program is not intended to be manually executed but wrapped by
"make local-travis". The program uses the env var TRAVIS_CONFIG to
know where to find the travis yaml file (the makefile will have it all
set for you).

We also use the http_proxy, https_proxy and no_proxy if they are set.
If we're executing a docker command that supports -e flag then these
variables are also handled.


dorileo:
@mdhorn I updated the PR adding a mechanism to automatically update the code coverage. If the user runs make check-coverage and has no unstaged changes the coverage database will be automatically updated with an added commit.

dorileo:
@mdhorn hopefully my last push :D

PR #76, opened 2018-09-26
post-install: add post-install hooks support

dorileo:
Introduce the initial support to post-install hooks. The code also
expands a set of pre-defined variables, for now the only exposed
variable is the chrootDir (the directory where the target system is
mounted).

### DIFF
--- a/tests/valid-minimal.yaml
+++ b/tests/valid-minimal.yaml
@@ -23,3 +23,14 @@ telemetry: false
 keyboard: us
 language: en_US.UTF-8
 kernel: kernel-native
+pre-install: [
+   {cmd: 'echo "running pre-install hook. dir: $chrootDir, chroot? $chrooted"'}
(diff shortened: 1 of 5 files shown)


dorileo:
@mdhorn added the pre-install hooks

tests/valid-minimal.yaml
```diff
@@ -23,3 +23,13 @@ telemetry: false
 keyboard: us
 language: en_US.UTF-8
 kernel: kernel-native
+pre-install: [
+   {cmd: 'echo "running pre-install hook form: $chrootDir"'}
```
mdhorn: is chrootDir set to "/" when chroot: false ?

dorileo: nope

dorileo:
@mdhorn updated, thanks.

PR #77, opened 2018-09-26
storage: add options field to block device structure

dorileo:
With this users may provide custom arbitrary mkfs.* options.


dorileo:
@mdhorn updated and also captured your sanity checks requests in the issue #82 

thanks...

PR #78, opened 2018-09-27
model: add Version field

dorileo:
The mass-installer use case for image generation requires to specify
the target system version. This patch allows the configuration of
an arbitrary version.


dorileo:
@mdhorn done. thanks.

PR #83, opened 2018-09-28
storage: add block-devices alias support

dorileo:
**This PR includes:**
- block device alias support - users may declare alias and use'em throughout the target media configuration;
- add image support - this is a special alias block device configuration, the controller will setup the loop device and expand the block device's name accordingly
- a small testsuite shuffle

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -7,6 +7,7 @@ package model
 import (
 	"io/ioutil"
 	"os"
+	"path/filepath"
 
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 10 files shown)


dorileo:
@mdhorn updated changing what you suggested. thanks.

model/model.go
```diff
@@ -26,7 +27,10 @@ import (
 // Also used by the Makefile for releases.
 // Default to the version of the program
 // but may be overridden for demo/documentation mode.
-var Version = "0.9.0"
+var (
```
mdhorn: Can you put this new variable on its own line?
Otherwise, the Makefile will fail for target 'tag'
https://github.com/clearlinux/clr-installer/blob/master/Makefile#L201

dorileo: Sure.

dorileo:
@mdhorn pushed new version fixing the control.Version. Thanks.

PR #89, opened 2018-10-02
args: rename --mirror to --swupd-mirror

dorileo:
Preventing future confusions (when we implement dnf support) rename
the --mirror argument to --swupd-mirror and start creating a namespace
for the arguments.

### DIFF
--- a/args/args.go
+++ b/args/args.go
@@ -114,8 +114,8 @@ func (args *Args) setCommandLineArgs() (err error) {
 		&args.ConfigFile, "config", "c", args.ConfigFile, "Installation configuration file",
 	)
 
-	flag.StringVarP(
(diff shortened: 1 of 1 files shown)


args/args.go
```diff
@@ -115,7 +115,7 @@ func (args *Args) setCommandLineArgs() (err error) {
 	)
 
 	flag.StringVarP(
-		&args.SwupdMirror, "mirror", "m", args.SwupdMirror, "Swupd Installation mirror URL",
+		&args.SwupdMirror, "swupd-mirror", "m", args.SwupdMirror, "Swupd Installation mirror URL",
```
mdhorn: Should we leave the short option name `-m`?
Maybe we should make it `-s` or drop it.

dorileo: had the same doubt, I'll drop it.

dorileo:
@mdhorn updated dropping the -m

PR #91, opened 2018-10-02
configuration and installer image changes

dorileo:
This PR proposes:
- replacing the mk-installer-image.sh script to a yaml descriptor (with its own post-install hook), though using installer to create the installer image - covering all the image creation aspects;
- support the use of both device file and image at the same time
- increased model's test coverage


dorileo:
@mdhorn updated adding the --block-device argument. Also updated the Readme.md file.

dorileo:
@mdhorn just found a couple of issues that will require my attention.

scripts/installer-image.yaml
```diff
@@ -0,0 +1,37 @@
...
+    mountpoint: /
+    size: "958M"
+    type: part
+
+bundles: [os-core, os-core-update, telemetrics, os-installer, openssh-server]
+telemetry: false
+keyboard: us
+language: en_US.UTF-8
+kernel: kernel-native
+
```
mdhorn: This config should add:
```
postReboot: false
postArchive: false
autoUpdate: false
```
then the --reboot false not needed on the command line
We do not want to store the log file and the configuration file in the target installer image
We do not want the image to update to the lastest.

mdhorn: @dorileo should we drop the '--reboot=false' from the Readme.md?

dorileo: sure

dorileo: done, thanks.

PR #94, opened 2018-10-03
Local travis for review

dorileo:
- add build-local-travis target to Makefile
- add --config flag to local-travis

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,9 +72,11 @@ build: gopath
 	go get -v ${GO_PACKAGE_PREFIX}/clr-installer
 	go install -v ${GO_PACKAGE_PREFIX}/clr-installer
 
-check-coverage: gopath
+build-local-travis: gopath
 	@go get -v ${GO_PACKAGE_PREFIX}/local-travis
 	@go install -v ${GO_PACKAGE_PREFIX}/local-travis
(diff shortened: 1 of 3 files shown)


dorileo:
@mdhorn seems to be passing now

dorileo:
@mdhorn fixed the issue with the makefile

Makefile
```diff
@@ -84,7 +84,8 @@ check: gopath
 	@# Ensure no temp files are left behind
 	@$(eval BEFORESHA = $(shell ls -art /tmp | sha512sum))
 	go test -cover ${GO_PACKAGE_PREFIX}/...
-	@if [ "$(BEFORESHA)" != "$$(ls -art /tmp | sha512sum)" ] ; then \
+	@$(eval AFTERSHA = $(shell ls -art /tmp | sha512sum))
```
mdhorn: Well, the problem here is that both the BEFORE and AFTER sha are calculated before any make targets are run since they are created with eval.
I ran into this in the original implementation and that's why I had to change it to be in the test.

dorileo: how come?

mdhorn:
@dorileo Looks to be stable now. Great find and fix!

PR #96, opened 2018-10-03
crypt: Ensure all state is cleaned up on error

lpereira:
Not all memory was cleaned up on error.  Also, avoid leaving a copy of
the password in plain text when returning from the Crypt() function.

Signed-off-by: Leandro Pereira <leandro.pereira@intel.com>

### DIFF
--- a/crypt/crypt.go
+++ b/crypt/crypt.go
@@ -25,7 +25,7 @@ const char dict[DICT_SIZE] =
 char *__salt_and_crypt(char prefix[3], char *pass, size_t size) {
   int i, fd;
   ssize_t randSize;
-  char rand[ENTROPY_SIZE], salt[SALT_SIZE], *enc;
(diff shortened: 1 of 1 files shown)


crypt/crypt.go
```diff
@@ -83,7 +85,10 @@ const (
 // the algorithm is not selectable/customizable.
 func Crypt(password string) (string, error) {
 	cPassword := C.CString(password)
-	defer C.free(unsafe.Pointer(cPassword))
+	defer func() {
+		C.memset(unsafe.Pointer(cPassword), 0, len(password))
```
otaviobp: According to error messages on travis build you can't use len(password) as a parameter.

mdhorn:
@otaviobp Not passing Travis is the reason it was never approved.

mdhorn:
The c code for encryption was not compatible with Gnome UI environment and has been replaced with native Go code.
https://github.com/clearlinux/clr-installer/pull/365
This PR is no longer applicable.

Issue #28, opened 2018-08-13
Improve Adduser Input data


mdhorn:
<a href="https://github.com/mdhorn"><img src="https://avatars.githubusercontent.com/mdhorn" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mdhorn](https://github.com/mdhorn)**
_Friday Jul 13, 2018 at 22:38 GMT_

----

- Request the user enter the password in two duplicate fields to ensure they know the password.
- Allow the user to enter their proper name (aka comment field)
- Allow the installer to set a password for the 'root' user

mdhorn:
<a href="https://github.com/mdhorn"><img src="https://avatars.githubusercontent.com/mdhorn" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mdhorn](https://github.com/mdhorn)**
_Friday Jul 13, 2018 at 22:39 GMT_

----

Testing of the Adduser function during install resulted in not being able to log into the account.
The account is created correctly, but the password is not matching the installed hash in /etc/shadow

PR #100, opened 2018-10-04
 Improve User account data entry

mdhorn:
Improve User account data entry

Fixes Issue: #28

### DIFF
--- a/tui/common.go
+++ b/tui/common.go
@@ -326,10 +326,14 @@ func (page *BasePage) setup(tui *Tui, id int, btns int, returnID int) {
 	// same as the default 'Cancel' button
 	page.window.OnKeyDown(func(ev clui.Event, data interface{}) bool {
 		if ev.Key == term.KeyEsc {
-			page.action = ActionCancelButton
-			page.GotoPage(returnID)
-			page.action = ActionNone
-			return true
+			if page.cancelBtn != nil {
+				page.cancelBtn.ProcessEvent(clui.Event{Type: clui.EventKey, Key: term.KeyEnter})
+			} else {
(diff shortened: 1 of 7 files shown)


dorileo:
There's an inconsistency, I add a user, no validation message is shown, but when I add a second user the validation messages are displayed when entering the "new user" form screen.

dorileo:
there's a padding space between user lines in the user list, is that intentional?

dorileo:
when editing a user if I press the "reset" button the "user name" field is selected and nothing else? is it intentional? wouldn't be the case of disabling this button if editing a user?

dorileo:
I have only these minor comments, otherwise LGTM.

mdhorn:
> There's an inconsistency, I add a user, no validation message is shown, but when I add a second user the validation messages are displayed when entering the "new user" form screen.

This

mdhorn:
> There's an inconsistency, I add a user, no validation message is shown, but when I add a second user the validation messages are displayed when entering the "new user" form screen.

This should be fixed now. Basically, the first time the code is run the default edit Title is an empty string and set it to empty string again did not trigger the OnChange code.

mdhorn:
> there's a padding space between user lines in the user list, is that intentional?

This was intentional to keep with spacing on other screen and provide visual space between user accounts.

mdhorn:
> when editing a user if I press the "reset" button the "user name" field is selected and nothing else? is it intentional? wouldn't be the case of disabling this button if editing a user?

The "reset" button was intended to put the User form back to the state when the screen was started. If adding a user, it would clear all the data; if modifying a user, it would restore the data to the starting state.

Issue #74, opened 2018-09-26
Enable Public Key for new user accounts


mdhorn:
All specifying a public ssh key when a new user account is created.

Instead of, or in addition to, a password for new users, allow for the use of a public key.

dorileo:
done.

PR #111, opened 2018-10-16
Enable Public Key for new user accounts

dorileo:
Fixes Issue: #74 

- Enable Public Key for new user accounts
- Allow users to be created without password

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -188,17 +188,19 @@ func (u *User) apply(rootDir string) error {
 		return errors.Wrap(err)
 	}
 
-	args = []string{
-		"chpasswd",
-		"--root",
-		rootDir,
-		"-e",
-	}
+	if u.Password != "" {
+		args = []string{
+			"chpasswd",
+			"--root",
+			rootDir,
+			"-e",
+		}
 
-	pwd := fmt.Sprintf("%s:%s", u.Login, u.Password)
+		pwd := fmt.Sprintf("%s:%s", u.Login, u.Password)
 
-	if err := cmd.PipeRunAndLog(pwd, args...); err != nil {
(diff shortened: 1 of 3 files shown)


user/user.go
```diff
@@ -186,17 +188,56 @@ func (u *User) apply(rootDir string) error {
...
-	if err := cmd.PipeRunAndLog(pwd, args...); err != nil {
-		return errors.Wrap(err)
+	return nil
+}
+
+func writeSSHKey(rootDir string, u *User) error {
+	dpath := filepath.Join(rootDir, "home", u.Login, ".ssh")
+	fpath := filepath.Join(dpath, "authorized_keys")
+
+	if err := utils.MkdirAll(dpath, 0755); err != nil {
```
mdhorn: These permissions are too permissive
```suggestion
	if err := utils.MkdirAll(dpath, 0700); err != nil {
```

user/user.go
```diff
@@ -186,17 +188,56 @@ func (u *User) apply(rootDir string) error {
...
+		return err
+	}
+
+	f, err := os.OpenFile(fpath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
```
mdhorn: These permissions are too permissive
```suggestion
	f, err := os.OpenFile(fpath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
```

mdhorn:
@dorileo After the permission changes, can you run a mass install with a real ssh key and make sure the created user can login with ssh.

dorileo:
@mdhorn updated the permission also change to adjust the file's owner. And yes, it's tested.

Thanks...

PR #116, opened 2018-10-22
Locales, timezone, and keyboard settings

mdhorn:
- Add error checking for valid values
- Fix TUI for corner-case of no locales, timezone, or keyboards
- Properly set the user choice on the target including adding required bundles

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -182,6 +182,10 @@ func (si *SystemInstall) Validate() error {
 		}
 	}
 
+	if si.Timezone == nil {
+		return errors.ValidationErrorf("Timezone not set")
+	}
+
 	if si.Keyboard == nil {
 		return errors.ValidationErrorf("Keyboard not set")
 	}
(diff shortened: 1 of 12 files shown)


model/model.go
```diff
@@ -182,6 +182,10 @@ func (si *SystemInstall) Validate() error {
 		}
 	}
 
+	if si.Timezone == nil {
```
dorileo: why are we considering timezone mandatory, if it's not provided then we should fallback to a default one - could be defined with the default configuration.

dorileo: you do what I'm suggesting with keyboard, just replicate it here maybe.

dorileo:
@mdhorn the PR looks good, just take a look at what I said about the timezone validation code.

Thank you.

model/model.go
mdhorn: If you look around line 250, we do default to the UTC if one is not defined. That check is a just a last ditch check or if code changes in the future to avoid a runtime nil error.

PR #117, opened 2018-10-23
Add open ssh server

mdhorn:
SSH Server should be in each install target by default.


dorileo:
@mdhorn you mention you're adding the bundle to bundles.json but I don't see such a change in the patch.

mdhorn:
@dorileo rebased on latest master.

mdhorn:
@dorileo Ready for review

PR #121, opened 2018-10-24
Images

dorileo:
- support legacy bios mode
- add aws image descriptor
- add azure image descriptor


dorileo:
added a set of new image descriptors.

PR #122, opened 2018-10-24
swupd: use --keepcache with "update" command

dorileo:
If we don't explicitly specify the --keepcache flag the previously
fetch content (specially when using --swupd-state flag) will be
cleaned up and all the content will be downloaded all over again.

### DIFF
--- a/swupd/swupd.go
+++ b/swupd/swupd.go
@@ -128,6 +128,7 @@ func (s *SoftwareUpdater) Update() error {
 	args := []string{
 		filepath.Join(s.rootDir, "/usr/bin/swupd"),
 		"update",
+		"--keepcache",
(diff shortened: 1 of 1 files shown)


otaviobp:
This commit makes sense to me, but be sure you are not creating a case where you are using the same state dir for different content_urls. This can prevent packs to be downloaded and will force full file downloads that are a lot slower.

For example if you have mix1 with version 10 based on clear-25780 and mix2 with version 10 based on clear-25790, when you install os-core from mix1 swupd will download pack-os-core-10.tar file. If you try later to use same state dir with mix2, swupd will need pack-os-core-10.tar file too, but those files are different. Swupd won't download the new pack and it will fail to check for hashes. As a fallback we use fullfiles, but this is far from the expected in this case.

PR #138, opened 2018-11-08
swupd: add new flags

dorileo:
Add the swupd's --format, --contenturl and --versionurl arguments.
These arguments are mapped with the --swupd prefix such as:
--swupd-format, --swupd-contenturl and --swupd-versionurl

And are used for constrained environments dealing with local installs,
specially with image generation in a network isolated environment.

### DIFF
--- a/swupd/swupd.go
+++ b/swupd/swupd.go
@@ -32,11 +32,12 @@ var (
 
 // SoftwareUpdater abstracts the swupd executable, environment and operations
 type SoftwareUpdater struct {
-	rootDir    string
-	stateDir   string
-	format     string
-	contentURL string
-	versionURL string
+	rootDir            string
+	stateDir           string
+	format             string
+	contentURL         string
+	versionURL         string
+	skipDiskSpaceCheck bool
 }
 
 // Bundle maps a map name and description with the actual checkbox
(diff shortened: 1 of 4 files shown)


swupd/swupd.go
```diff
@@ -66,6 +94,9 @@ func (s *SoftwareUpdater) Verify(version string, mirror string) error {
 		"swupd",
 		"verify",
 	}
+
+	args = s.setExtraFlags(args)
+
 	if mirror != "" {
 		args = append(args, fmt.Sprintf("--url=%s", mirror))
```
mdhorn: Is there any conflict in settings --versionurl --contenturl and --url?

dorileo: nope

swupd/swupd.go
```diff
@@ -104,9 +135,14 @@ func (s *SoftwareUpdater) Verify(version string, mirror string) error {
 		"swupd",
 		"bundle-add",
 		"--skip-diskspace-check",
```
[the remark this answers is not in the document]
dorileo: good idea but I would rather start setting it to true by default, what do you think? I see most of the cases we want to use it, but we still have the chance to disable anyway.

dorileo:
@mdhorn added a new patch to make --skip-diskspace-check optional. But still setting it to true by default.

PR #152, opened 2018-11-09
images: reduce the swap partition

dorileo:
Reduce the swap partition to 32M with the live-desktop image.

### DIFF
--- a/scripts/live-desktop.yaml
+++ b/scripts/live-desktop.yaml
@@ -18,12 +18,12 @@ targetMedia:
     type: part
   - name: ${bdevice}2
     fstype: swap
-    size: "256M"
+    size: "32M"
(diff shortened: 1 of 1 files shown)


mdhorn:
Given this is a Live Image, and has a GUI desktop, should we really be reducing the swap to such a small value?

dorileo:
@mdhorn personally I don't see much difference between 150M and 32M for swap partition in the virtual memory side of the things, if you're really a heavy swap user candidate 150M will also not save you either

mdhorn:
@dorileo I think our swap is already too small. Do we have any study or data that shows we should have such a small swap?

dorileo:
I think @fenrus75 could give us more data about this.

PR #153, opened 2018-11-12
live-desktop: install custom default settings

dorileo:
The live-desktop has a custom set of default bundles, this patch
makes sure we install the proper default settings on post install
script.


mdhorn:
Will change the file under /usr still allow swupd to not complain?

dorileo:
@mdhorn swupd auto-update is disabled by default on images, however, do you think we should add one more level in the file loading hierarchy?

dorileo:
maybe somewhere in /var/lib/clr-installer (somewhere swupd will never touch with a swupd verify) ?

mdhorn:
@dorileo Yes, I'm afraid a swupd verify might fail or verify --fix might remove it.

dorileo:
@mdhorn ok, I'm finishing a patch to implement that.

dorileo:
@mdhorn done

mdhorn:
@dorileo We should also be consistent in the bundles between the yaml files. In some we list both desktop and desktop-autostart, but in the other only desktop-autostart.

dorileo:
@mdhorn you're not seeing the list because it requires a change in the clr-installer (see patch 7892a4c5c634acae2e753c370a4c6f67a0706f80 in this PR). When you build a new image it installs clr-installer from the bundle clr-installer and since this patch has not been merged yet the release was not made and the package/bundle was not updated you'll not see the changes in the generated image, unless you manually copy the resulting binary (you can use the post install script to do that for you).

dorileo:
@mdhorn I'll make sure to keep the desktop-autostart vs desktop usage consistent across the multiple lists (I'll update this PR ASAP).

dorileo:
@mdhorn done the desktop bundle consistency changes. thanks.

PR #155, opened 2018-11-13
storage: bug fix for calculating image size

mdhorn:
- Calculate size either at the parent level or as a sum of the children.
- Add error checking and new test cases.


mdhorn:
@dorileo Corrected. Please review.

PR #173, opened 2018-11-28
Check for root early

mdhorn:
Change the timing where we check for the root user.

Fixes: 171

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -456,11 +456,9 @@ func listBlockDevices(userDefined []*BlockDevice) ([]*BlockDevice, error) {
 		return nil, err
 	}
 
-	if err = utils.VerifyRootUser(); err == nil {
-		for _, bd := range bds {
-			if err = bd.PartProbe(); err != nil {
-				return nil, err
-			}
+	for _, bd := range bds {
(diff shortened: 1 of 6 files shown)


storage/storage.go
```diff
@@ -456,7 +456,7 @@ func listBlockDevices(userDefined []*BlockDevice) ([]*BlockDevice, error) {
 		return nil, err
 	}
 
-	if err = utils.VerifyRootUser(); err == nil {
+	if errString := utils.VerifyRootUser(); errString == "" {
```
dorileo: With the early check, do we need this check at all?

mdhorn: Maybe not? I was concerned there was some other code flow which used storage before the first check. You think I should drop it?

dorileo: Yep, please drop it.

mdhorn: Dropping. First fall out is one of the storage tests is now failing when not run as root. Updating...

mdhorn: @dorileo Now now the coverage numbers of storage have dropped. :(

mdhorn: @dorileo Updates made as we discussed offline.
New changes are pushed and testing now.

PR #177, opened 2018-11-30
Make make work without git repo

mdhorn:
Production builds are made from github source archives. As there
is no repository, git commands will not be available for builds.


mdhorn:
@dorileo Please review as this is needed to unblock the pending release.

Issue #178, opened 2018-12-03
Create a yaml configuration export


mrkz:
Please provide a way to generate a `*.yml` file once a user configured the installation parameters using the `TUI`.

There is a way to automatically install the OS using an existing `*.yml`, but the documentation for generating a `yaml` is lacking, so a user-fiendly method to create a yaml file for future installs using the TUI would be much appreciated.

This would also help for users to later report issues to clr-installer developers, as it would be much easier for users to share their setup when affected by a bug (my case).

**Expected behavior**
 - User starts the `clr-installer`
 - User configure their desired options such as disk partitions schema, users, bundles, etc
 - Previous to hit the `install` button, an option (button?) to save current configuration as a yaml file is available

dorileo:
@mdhorn not sure :(

mrkz:
I'd prefer the option of a `save to yaml` button, as I wouldn't know if I were to save the config (probably the requirement pops out at the middle of the set up).

The `--save-config` option sounds appealing, but please do consider adding the button as well.

AntonioMeireles:
(thinking loud) what about doing what `anaconda` has being doing for ages ... place *by default* the `yaml` on the newly assembled root image and have a flag / button to disable it?. I'd expect this to be a behaviour that most newcomers would expect and be accustomed to ...

mdhorn:
@AntonioMeireles  The installer does place the YAML file on the newly installed system along with the installation log file.

What @mrkz was asking for was a way to use the TUI to generated the same YAML without doing the actual installation.

AntonioMeireles:
@mdhorn many thanks for the clarification! (just following this repo - not  consuming new installer yet for my tooling). sorry for the noise!

mrkz:
@AntonioMeireles yeah, what I'm asking here basically it's for a way to generate the yaml file for automated installs, populating this yaml via the installer options, later on saving the file instead or previous to continue with the installation.

AntonioMeireles:
understood, thanxs! - a `--dry-run` flag / mode. (my suggestion for its naming)

mdhorn:
Instead of having a command line option, since the "normal" use case is needed to save once the TUI is already running, instead implemented a new Advance option which saves the current configuration to a YAML file.

@mrkz Can you see if PR https://github.com/clearlinux/clr-installer/pull/183 meets your needs.

mrkz:
[...]

PR #183, opened 2018-12-06
Allow TUI user to save configuration file w/o Install

mdhorn:
Fixes Issue: #178 

- Add Advanced Save Config option
- Add date/time stamp to the configuration file


mdhorn:
@dorileo So I literally copy the other popup window code to make this new file.
If we make these changes here, then the other copies should be updated in the same way.

mdhorn:
@dorileo I've fixed the bug we discussed earlier and added a save of the configuration file prior to starting the install for debug/bug reporting. Please re-review.

PR #202, opened 2019-01-25
Introduction of the GUI

ikeyd:
This GUI has been implemented in GTK+ and is very much a work in
progress. So far, we have a number of initial pages implemented:

 - Disk Configuration
 - Bundles
 - Keyboard
 - Language
 - Timezone
 - Telemetry

Note, some of the pages are quite ugly and basic and will need
some improvements. Additionally, the disk config page is currently
very crude, only supporting automatic installation mode.

Additionally, the use of pkexec will ignore environmental variables
for the proxy, thus, proxy support is a bust in the GUI installer.
Some OS-level work is expected to resolve this magically anyway.

There are some ugly hardcoded elements, such as relying on a system
wallpaper file for CSS, and the cooked-in CSS vs an on-disk .css file
that designers can edit and tweak locally. Note that error handling
is incredibly immature right now, resulting in a panic. Realistically
we need to show a "success" or "failure" page within a stack on the
install page to let the user know what happened, and offer the choice
of a reboot.

As a final note, the UX is still not rock solid, and will need
a design-develop feedback loop to improve. This commit represents
my initial handoff for others to jump in and vastly improve the
GUI.


ikeyd:
Note the travis test will need modifying

dorileo:
I suggest improving this patchset granularity, adding the basic bits incrementally so we can start merging the code even it doesn't fully implement the features but it should be functional, I shouldn't break things with that you don't need to have all the huge amount of code sitting on your WIP branch much longer than you need - rebasing such big chunck of code could easily become a nightmare.

mdhorn:
@ikeyd This is merged into the dev-gui branch were we will continue the GUI work until the first beta is ready at which time we will merge dev-gui into master.

Issue #203, opened 2019-01-25
Use LTS Kernel when installing in VirtualBox


mdhorn:
When installing Clear Linux OS in VirtualBox, the installer should recognize this installation environment and automatically select the LTS kernel for the install.

mdhorn:
@fenrus75 Should we also be defaulting to the LTS kernel for Parallels and VMware environment too?

The current PR is only changing the default kernel from Native to LTS for interactive installs.

fenrus75:
nope; the virtual box is special in that we only support their drivers in
the LTS kernel package .... parallels/vmware are not like that

PR #204, opened 2019-01-28
Use LTS Kernel for VirtualBox Installs

mdhorn:
Fixes Issue: #203

- If running interactively and in a VirtualBox environment force the default to be the kernel-lts.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -16,6 +16,8 @@ import (
 	"syscall"
 	"unsafe"
 
+	"github.com/digitalocean/go-smbios/smbios"
+
 	"github.com/clearlinux/clr-installer/errors"
 )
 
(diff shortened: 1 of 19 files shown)


utils/utils.go
```diff
@@ -181,3 +183,44 @@ func ExpandVariables(vars map[string]string, str string) string {
...
+// from within VirtualBox
+// Attempt to parse the System Management BIOS (SMBIOS) and
+// Desktop Management Interface (DMI) to determine if we are
+// executing inside a VirtualBox. Ignoring error conditions and
+// assuming we are not VirtualBox.
+func IsVirtualBox() bool {
+	virtualBox := false
+
+	// Find SMBIOS data in operating system-specific location.
+	rc, _, err := smbios.Stream()
```
dorileo: are you sure there isn't a case where rc will not require a call for rc.Close() even if we got an err? I would suggest moving the deferred rc.Close() up here.

dorileo:
LGTM - just check the rc.Close() requirements, if will never need to rc.Close() when an error is returned then don't even ask, go ahead and merge.

PR #207, opened 2019-01-31
docs: Readme.md updates

pdxjohnny:
* Removed 'n' from clearnlinux, should be clearlinux


mdhorn:
@pdxjohnny thanks for the fix

Issue #209, opened 2019-02-05
Adding new user crashes the installer


mdhorn:
**Describe the bug**
Sometimes the installer crashes when adding a new user.

**To Reproduce**
1. Download the installer image for Clear Linux OS 27000
2. Run the installer image
3. Ensure auto-update is left enable (seeing this problem 27000 --> 27600)
4. Add a user under the Advanced Options
5. Start the installer
6. See error 
-- From the clr-installer.log in /
`2019/02/04 17:40:48 [DBG] *** stack smashing detected ***: <unknown> terminated
`

**Expected behavior**
The installer should complete without error and the target media should have the requested user.

**Environment (please complete the following information):**
 - Clear Linux OS Version: 27000 --> 27600
 - Installer Version 1.1.4

**Additional context**

PR #210, opened 2019-02-05
Chroot first, then run useradd

mdhorn:
Fixes Issue: #209

- Prefix account commands with chroot to ensure using the right version of libraries


gtkramer:
+1

Issue #193, opened 2019-01-10
Ister JSON to clr-installer YAML converter


mdhorn:
Need to create a tool, or functionality in clr-installer to read in Ister.py JSON files and generate the equivalent YAML file.  

If it could be built-into clr-installer to use the Ister JSON and output the YAML as a side-effect would be ideal, but maybe not reasonable.

mdhorn:
This was release in 1.1.7

PR #226, opened 2019-02-19
Convert ister json config to clr-installer yaml config

reaganlo:
Fixes Issue: #193 

 - The new -j flag detects ister json config format and converts it to yaml.
- The existing -c flag also detects ister json config format, converts it to yaml and continues as usual.
- Handles the ister "rest" option for physical targets
- Handles multiple targets


model/model_ister.go
```diff
@@ -0,0 +1,327 @@
...
+		si.PostInstall = append(si.PostInstall, &pi)
+	}
+
+	si.LegacyBios = ic.LegacyBios // Set LegacyBios
+	si.SwupdMirror = ic.MirrorURL // Set SwupdMirror
+
+	si.HTTPSProxy = ic.HTTPSProxy // Set HTTPSProxy
+	if si.HTTPSProxy == "" {
+		si.HTTPSProxy = ic.HTTPProxy
+		fmt.Println("WARNING: Mapping HTTPProxy in json to HTTPSProxy in yaml")
```
mdhorn: Probably should be using log.Warning here

model/model_ister.go
```diff
@@ -0,0 +1,327 @@
...
+	} else {
+		fmt.Println("WARNING: Skipping HTTPProxy mapping")
```
mdhorn: Probably should be using log.Warning here

model/model_ister.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	// Hardcoding the missing required fields
+	si.Telemetry = &telemetry.Telemetry{Enabled: false}              // Set Telemetry
+	si.Keyboard = &keyboard.Keymap{Code: "us"}                       // Set Keyboard
```
mdhorn: Can we switch the default keyboard to keyboard.DefaultKeyboard

model/model_ister.go
```diff
@@ -0,0 +1,327 @@
...
+	si.Language = &language.Language{Code: language.DefaultLanguage} // Set Language
+
+	if ic.VersionURL != "" {
+		fmt.Println("WARNING: Skipping VersionURL mapping as it not supported in clr-installer config")
```
mdhorn: Probably should be using log.Warning here

mdhorn: Did we find any existing use cases where the JSON includes a versionURL?

storage/ops.go
```diff
@@ -398,11 +398,21 @@ func commonMakeFsCommand(bd *BlockDevice, args []string) ([]string, error) {
 }
 
 func commonMakePartCommand(bd *BlockDevice, start uint64, end uint64) (string, error) {
-	args := []string{
-		"mkpart",
-		bd.MountPoint,
-		fmt.Sprintf("%dM", start),
-		fmt.Sprintf("%dM", end),
+	var args []string
+	if end == 0 {
```
mdhorn:
```suggestion
     var trueEnd int64 = int64(end)

    if end == 0 {
        trueEnd = -1
    }

    args = []string{
        "mkpart",
        bd.MountPoint,
        fmt.Sprintf("%dM", start),
        fmt.Sprintf("%dM", trueEnd),
    }
```

clr-installer/main.go
```diff
@@ -176,6 +189,13 @@ func main() {
 		defer func() { _ = os.Remove(cf) }()
 	}
 
+	if filepath.Ext(cf) == ".json" {
+		cf, err = model.JSONtoYAMLConfig(cf)
+		if err != nil {
+			fatal(err)
+		}
+	}
```
mdhorn: We should add log.Info() message here that the JSON was automatically converted to YAML.

mdhorn:
@reaganlo I still want to run this a few times myself before I approve the final code.

clr-installer/main.go
```diff
@@ -141,6 +142,18 @@ func main() {
...
 	}
 
+	if options.ConvertConfigFile != "" {
+		if filepath.Ext(options.ConvertConfigFile) == ".json" {
+			_, err = model.JSONtoYAMLConfig(options.ConvertConfigFile)
+			if err != nil {
+				fatal(err)
+			}
+		} else {
+			fmt.Println("Config file is not in JSON format")
```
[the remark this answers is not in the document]
reaganlo: Added a fatal error with suggested message.

model/model_ister.go
```diff
@@ -0,0 +1,327 @@
+// Copyright © 2018 Intel Corporation
```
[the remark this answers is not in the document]
reaganlo: Done

model/model_ister.go
```diff
@@ -0,0 +1,327 @@
...
+
+	"github.com/clearlinux/clr-installer/kernel"
+	"github.com/clearlinux/clr-installer/keyboard"
+	"github.com/clearlinux/clr-installer/log"
+	"github.com/clearlinux/clr-installer/storage"
+	"github.com/clearlinux/clr-installer/telemetry"
+	"github.com/clearlinux/clr-installer/user"
+)
+
+// ImageConfig represents the  install configuration used  in the "ister" app
```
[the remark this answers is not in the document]
reaganlo: Done

model/model_ister.go
```diff
@@ -0,0 +1,327 @@
...
+//
+// SPDX-License-Identifier: GPL-3.0-only
+
+package model
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/clearlinux/clr-installer/language"
+	"github.com/clearlinux/clr-installer/network"
```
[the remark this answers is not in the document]
reaganlo: Done

model/model_ister.go
```diff
@@ -0,0 +1,327 @@
...
+	"github.com/clearlinux/clr-installer/kernel"
+	"github.com/clearlinux/clr-installer/keyboard"
+	"github.com/clearlinux/clr-installer/log"
+	"github.com/clearlinux/clr-installer/storage"
+	"github.com/clearlinux/clr-installer/telemetry"
+	"github.com/clearlinux/clr-installer/user"
+)
+
+// ImageConfig represents the  install configuration used  in the "ister" app
+type ImageConfig struct {
```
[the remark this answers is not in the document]
reaganlo: Done

model/model_ister.go
reaganlo: Done. But since log.Warning does not print to screen, kept the existing fmt.Println

model/model_ister.go
reaganlo: Done. But since log.Warning does not print to screen, kept the existing fmt.Println

model/model_ister.go
reaganlo: Done

model/model_ister.go
reaganlo: Done. But since log.Warning does not print to screen, kept the existing fmt.Println.
No, I have not come across an ister conf with VersionURL set.

storage/ops.go
reaganlo: `var trueEnd int64 = int64(end)` gives a lint warning. So did it this way instead:
```
	var trueEnd int64
	if end == 0 {
		trueEnd = -1
	} else {
		trueEnd = int64(end)
	}
```

clr-installer/main.go
reaganlo: Since the JSONtoYAMLConfig() is called twice in the code, added the log.Info() message inside this function along with the corresponding fmt.Println message.

mdhorn:
@reaganlo Sorry for the delay in testing.
One thing I'd like to see changed is when mapping the virtual disk,
```
       "DestinationType": "virtual",
       "PartitionLayout": [
               {
                       "disk": "clear.img",

```
instead of generating this alias:

`block-devices: [{name: clear.img, file: clear.img}]
` 
Please remove the extentension from the alias name:
`block-devices: [{name: clear, file: clear.img}]`

model/model_ister.go
```diff
@@ -0,0 +1,332 @@
...
+
+		partitions, ok := disks[curr.Disk]
+		if !ok {
+			sa := StorageAlias{}
+			bd := storage.BlockDevice{}
+			switch ic.DestinationType {
+			case "virtual":
+				bd.Name = curr.Disk
+				bd.Type = storage.BlockDeviceTypeLoop
+				sa.Name = curr.Disk
```
mdhorn:
```suggestion
				sa.Name = filepath.Base(curr.Disk)
```

mdhorn:
Could we look at changing the naming of the converted files?

Ideally, I'd like to see:
- Use the same filepath.Base(json_file) + .".yaml"
  - If the yaml file already exists,
    - then rename current yaml file to filepath.Base(json_file) + datestamp (from stat of file) + .".yaml"
This means the user always gets the latest version in the standard file name, but backups of previously converted files are not lost.

Bonus Point: Add a --no-backups flag that always just overwrites and does not make backups for those bold enough to risk losing work (and like a clean directory)

PR #229, opened 2019-02-20
storage: Fix partition misalignment issue

chenming1986:
In the cases that 1MiB is the optimal alignment, specifying 0M as the
start position to the mkpart command will lead parted to create
misaligned partitions even when the option -a optimal is set. It could
cause dramatic performance degradation to I/O intensive workloads. Set
the start position to 1M so that the first partition can be properly
aligned and so are the other partitions.

Signed-off-by: Ming Chen <ming.a.chen@intel.com>

Fixes Issue: #

-
-
-

### DIFF
--- a/storage/ops.go
+++ b/storage/ops.go
@@ -194,7 +194,7 @@ func (bd *BlockDevice) WritePartitionTable(legacyBios bool) error {
 		"--script",
 	}
 
-	var start uint64
+	var start uint64 = 1
 	bootPartition := -1
 	bootStyle := "boot"
 	guids := map[int]string{}


chenming1986:
Tested with a Samsung 970 EVO 256G NVMe and Clear Linux 29710 and auto partition was used during OS installation.

Without the commit, parted align-check optimal reported all 3 partitions were not aligned. fdisk -l output:
Device          Start       End   Sectors   Size Type
/dev/nvme0n1p1     34    292968    292935   143M EFI System
/dev/nvme0n1p2 292969    792968    500000 244.1M Linux swap
/dev/nvme0n1p3 792969 465771484 464978516 221.7G Linux root (x86-64)

With the commit, parted align-check optimal reported all 3 partitions were aligned. fdisk -l output:
Device          Start       End   Sectors   Size Type
/dev/nvme0n1p1   2048    294911    292864   143M EFI System
/dev/nvme0n1p2 294912    794623    499712   244M Linux swap
/dev/nvme0n1p3 794624 465772543 464977920 221.7G Linux root (x86-64)

mdhorn:
This was replaced with #240 which has already been merged.
Thank you.

PR #240, opened 2019-02-27
Ensure proper partition alignment

mdhorn:
- Ensure optimized partition are created

### DIFF
--- a/storage/ops.go
+++ b/storage/ops.go
@@ -25,7 +25,7 @@ import (
 type blockDeviceOps struct {
 	makeFsCommand   func(bd *BlockDevice, args []string) ([]string, error)
 	makeFsArgs      []string
-	makePartCommand func(bd *BlockDevice, start uint64, end uint64) (string, error)
+	makePartCommand func(bd *BlockDevice) (string, error)
 }
 
 var (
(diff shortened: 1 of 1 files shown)


storage/ops.go
```diff
@@ -461,32 +488,28 @@ func swapMakeFsCommand(bd *BlockDevice, args []string) ([]string, error) {
 	return cmd, nil
 }
 
-func swapMakePartCommand(bd *BlockDevice, start uint64, end uint64) (string, error) {
+func swapMakePartCommand(bd *BlockDevice) (string, error) {
 	partName := "linux-swap"
 
 	if bd.FsType == "swap" && bd.Type == BlockDeviceTypeCrypt {
-		mapped := fmt.Sprintf("eswap%d", start)
+		mapped := fmt.Sprintf("eswap-%s", bd.Name)
```
reaganlo: Was this a bug fix?

storage/ops.go
```diff
@@ -192,6 +207,7 @@ func (bd *BlockDevice) WritePartitionTable(legacyBios bool) error {
 		"optimal",
 		bd.GetDeviceFile(),
 		"--script",
+		"--",
```
reaganlo: What is this change fixing?

storage/ops.go
mdhorn: start is no longer available to make the eswap mapped name unique, so switching to using the device name.

storage/ops.go
mdhorn: The "--" is required to tell parted that there are no more options, else the use of '-1' for the script is treated as a command line option.

storage/ops.go
mdhorn: Here is what the change will do to the swap name:
```
# lsblk
NAME                                        MAJ:MIN RM  SIZE RO TYPE  MOUNTPOINT
sr0                                          11:0    1 1024M  0 rom
vda                                         253:0    0    8G  0 disk
├─vda1                                      253:1    0  142M  0 part
├─vda2                                      253:2    0  244M  0 part
│ └─eswap-vdb2                              252:1    0  243M  0 crypt [SWAP]
└─vda3                                      253:3    0  7.3G  0 part
  └─luks-1d955539-b532-4bcb-870f-37a0c0992696
                                            252:0    0  7.3G  0 crypt /
```

Previously it would have been eswap<SIZE>, so this naming is probably better and still unique.

reaganlo: cool.. Thanks for the info.. 
Curious..shouldn't it be "eswap-vd**a**2"?

mdhorn: The name will be mapped on the image that is running the installer; in this case, vda would have been the source media, and vdb would be the target. It just need to be unique.

PR #253, opened 2019-03-06
Fix desktop image size for USB media

mdhorn:
- Reduce the live desktop beta image size to ensure it fits on an 8GB USB media


mdhorn:
> Looks good. But how did we come up with this specific size - "7800000000"?

https://github.com/clearlinux/distribution/issues/503

The file used for the published image was already updated, this is to maintain consistency.
We should probably rename this file in the future so it is apparent this version is not used for publication, but for testing and as an example.

PR #259, opened 2019-03-08
Download User Messages

mdhorn:
- Download Pre and Post Installer messages
TODO: Display the messages as pop-ups

### DIFF
--- a/tui/install.go
+++ b/tui/install.go
@@ -7,11 +7,12 @@ package tui
 import (
 	"time"
 
-	"github.com/clearlinux/clr-installer/controller"
-	"github.com/clearlinux/clr-installer/progress"
-
 	"github.com/VladimirMarkelov/clui"
 	term "github.com/nsf/termbox-go"
+
+	"github.com/clearlinux/clr-installer/controller"
+	"github.com/clearlinux/clr-installer/network"
+	"github.com/clearlinux/clr-installer/progress"
 )
 
 // InstallPage is the Page implementation for installation progress page, it also implements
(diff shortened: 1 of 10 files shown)


tui/install.go
```diff
@@ -25,6 +30,26 @@ type InstallPage struct {
...
+
+// Messenger is an array of language string used to create
+// an end user message for display
+type Messenger struct {
+	Messages []*LangString `yaml:"install-msg,omitempty"`
+}
+
+const (
+	//installDataURLBase = "https://cdn.download.clearlinux.org/releases/28090/clear/config/image/.data"
+	installDataURLBase = "http://mhorn-linux.jf.intel.com/pre-release/.data/"
```
reaganlo: "http://mhorn-linux.jf.intel.com/pre-release/.data/"

tui/install.go
```diff
@@ -138,3 +164,46 @@ func newInstallPage(tui *Tui) (Page, error) {
...
+		return
+	}
+
+	var lines []string
+	found := false
+	for _, message := range result.Messages {
+		line := strings.TrimSpace(message.Default)
+		if len(line) > 0 {
+			// We found at least one line with non-whitespace
+			found = true
```
reaganlo: If we are not skipping empty lines why do we need this flag "found"?

mdhorn: This is fixed

tui/install.go
mdhorn: Fixed this to the proper URL, not a test.

PR #262, opened 2019-03-11
Fixes `make lint` timeout in Travis

reaganlo:
Fixes timeout issues in Travis caused due to the increase in `make lint` time because of the newly added gtk packages.


mdhorn:
This changes was pulled into #265

PR #265, opened 2019-03-12
Exclude vendor from linting

mdhorn:
- Exclude the vendor directory from lint commands
- Added new target for lint including vendor for release testing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,19 +41,19 @@ before_script:
     - docker exec -it clear-test bash -c "journalctl|cat"
 
 script:
-    - docker exec -it clear-test bash -c "cd /travis ; make dist-clean"
-    - docker exec -it clear-test bash -c "cd /travis ; make"
-    - docker exec -it clear-test bash -c "cd /travis ; PATH=${PATH}:${GOPATH}:/go/bin make vendor-check"
-    - docker exec -it clear-test bash -c "cd /travis ; make lint"
-    - travis_retry docker exec -it clear-test bash -c "cd /travis ; make check"
-    - docker exec -it clear-test bash -c "cd /travis ; make clean"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -212,7 +212,29 @@ lint: build install-linters gopath
 	@rm -rf ${LOCAL_GOPATH}/src/${GO_PACKAGE_PREFIX}/vendor
 	@cp -af vendor/* ${LOCAL_GOPATH}/src/
 	@go build -race github.com/clearlinux/clr-installer/...
-	@gometalinter.v2 --deadline=10m --tests --vendor \
+	@gometalinter.v2 --deadline=10m --tests --vendor --skip=vendor \
+	--exclude=vendor --disable-all \
```
reaganlo: Do we need this `--exclude=vendor --disable-all \` if we are already skipping vendor `--skip=vendor`?

mdhorn: exlcude Exclude messages matching these regular expressions
but skip, Skip directories with this name when expanding '...'.

So the skip should shorten the processing time by never running on the vendor code .

.travis.yml
```diff
@@ -44,7 +44,7 @@ script:
     - docker exec -it clear-test bash -c "cd /travis ; make dist-clean"
     - docker exec -it clear-test bash -c "cd /travis ; make"
     - docker exec -it clear-test bash -c "cd /travis ; PATH=${PATH}:${GOPATH}:/go/bin make vendor-check"
-    - docker exec -it clear-test bash -c "cd /travis ; make lint"
+    - travis_wait 30 sleep infinity & docker exec -it clear-test bash -c "cd /travis ; make lint"
```
reaganlo: So we are keeping lint deadline as 20m and travis_wait as 30m? Just checking if this is intentional.

mdhorn: No, they should be consistent.

Issue #227, opened 2019-02-19
Make All Bootable images Hybrid UEFI GPT + BIOS GPT/MBR


mdhorn:
**Is your feature request related to a problem? Please describe.**
Currently, all of the published images are created as GPT UEFI only images with the exception of the special case ISO installer image which is ONLY MBR BIOS.

**Describe the solution you'd like**
The images would be a lot more user-friendly if they all support booting either UEFI GTP or BIOS GPT/MBR. This would eliminate a lot of user initial confusion on how to boot the Clear Linux OS installer images.

mdhorn:
This feature has been available since the 2.0.0 release.

PR #273, opened 2019-04-01
Add ISO output support

evil-brain:
Fixes Issue: #227

- Add `iosutils` package which implements MakeIso() for ISO creation
- Add swupd flag to NOT install bundles beyond os-core
- Add the hook to clr-installer to build ISO files if requested.

ISO generation is based on the script `create-iso.sh`.

Questions:
 * What do I do with files that are needed for execution and need to be available on the filesystem?
   * Examples are: `isoutils/exclude, isoutils/initrd_init_template, isoutils/isolinux.cfg.template`, and both *.c32 files

 * To support a menu on Legacy systems, we need the two files `libutil.c32 and menu.c32`. These files do not exist in Clear Linux at all. The files are currently from a package in Ubuntu. What should we do?

 * There's now more prerequisites for clr-installer (but only if you intend to make ISO files). These programs are `mksquashfs, cpio, mkfs.fat, xorriso` and the bundle without executables is `bootloader`(for isolinux files). This list may not be complete.

 * I could probably move more things from RunAndLog to Go native - cp and umount for example. Suggestions?

* There are many image types that cannot have an ISO made from them at the moment, because there's a hard requirement on the image supporting EFI and the EFI partition must be mounted at '/boot' in relation to the new image's filesystem base. Any suggestions on how to report to the user/deal with this better?

* The size of the ramfs (after boot) is currently hardcoded in the initrd_init_template (line 139), should this be a config option in the YAML/somewhere? Is 256M enough?

### DIFF
--- a/isoutils/isoutils.go
+++ b/isoutils/isoutils.go
@@ -1,0 +1,574 @@
+// Copyright © 2019 Intel Corporation
+//
+// SPDX-License-Identifier: GPL-3.0-only
+
+package isoutils
+
+import (
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"text/template"
+
+	"github.com/clearlinux/clr-installer/args"
+	"github.com/clearlinux/clr-installer/cmd"
+	"github.com/clearlinux/clr-installer/log"
+	"github.com/clearlinux/clr-installer/model"
+	"github.com/clearlinux/clr-installer/progress"
+	"github.com/clearlinux/clr-installer/swupd"
+	"github.com/clearlinux/clr-installer/utils"
+)
+
+type paths int
+
+const (
+	clrEfi    paths = iota + 0
+	clrImgEfi       // Location to mount the EFI partition in the passed-in img file
+	clrInitrd
+	clrRootfs
+	clrCdroot
+)
+
+var (
+	tmpPaths = make([]string, 5)
+)
+
+func mkTmpDirs() error {
+	msg := "Creating directory trees"
+	prg := progress.NewLoop(msg)
+	log.Info(msg)
+	var err error
+
+	tmpPaths[clrEfi], err = ioutil.TempDir("", "clrEfi-")
+	if err != nil {
+		prg.Failure()
+		return err
+	}
+	tmpPaths[clrInitrd], err = ioutil.TempDir("", "clrInitrd-")
+	if err != nil {
+		prg.Failure()
+		return err
+	}
+	tmpPaths[clrCdroot], err = ioutil.TempDir("", "clrCdroot-")
+	if err != nil {
+		prg.Failure()
+		return err
+	}
+
+	/* Create specific directories for the new cd's root */
+	for _, d := range []string{
+		tmpPaths[clrCdroot] + "/isolinux",
+		tmpPaths[clrCdroot] + "/EFI",
+		tmpPaths[clrCdroot] + "/images",
+		tmpPaths[clrCdroot] + "/kernel",
+	} {
+		if _, err := os.Stat(d); os.IsNotExist(err) {
+			err = os.Mkdir(d, os.ModePerm)
+			if err != nil {
+				prg.Failure()
+				return err
+			}
+		}
+	}
+
+	prg.Success()
+	return err
+}
+
+func mkRootfs() error {
+	msg := "making rootfs squashfs"
+	prg := progress.NewLoop(msg)
+	log.Info(msg)
+
+	/* TODO: This takes a long time to run, it'd be nice to see it's output as it's running */
+	args := []string{
+		"mksquashfs",
+		tmpPaths[clrRootfs],
+		tmpPaths[clrCdroot] + "/images/rootfs.img",
+		"-b",
+		"131072",
+		"-comp",
+		"gzip",
+		"-e",
+		"boot/",
+		"-e",
+		"proc/",
+		"-e",
+		"sys/",
+		"-e",
+		"dev/",
+		"-e",
+		"run/",
+	}
+	err := cmd.RunAndLog(args...)
+	if err != nil {
+		prg.Failure()
+		return err
+	}
+	prg.Success()
+	return err
+}
+
+func mkInitrd(version string, model *model.SystemInstall) error {
+	msg := "Installing the base system for initrd"
+	prg := progress.NewLoop(msg)
+	log.Info(msg)
+
+	var err error
+	options := args.Args{
+		SwupdMirror:             model.SwupdMirror,
+		SwupdStateDir:           tmpPaths[clrInitrd] + "/var/lib/swupd/",
+		SwupdStateClean:         true,
+		SwupdFormat:             "staging",
+		SwupdSkipDiskSpaceCheck: true,
+	}
+	sw := swupd.New(tmpPaths[clrInitrd], options)
+
+	/* Should install the overridden CoreBundles above (eg. os-core only) */
+	if err := sw.Verify(version, model.SwupdMirror, true); err != nil {
+		prg.Failure()
+		return err
+	}
+
+	prg.Success()
+	return err
+}
+
+func mkInitrdInitScript(templatePath string) error {
+	msg := "Creating and installing init script to initrd"
+	prg := progress.NewLoop(msg)
+	log.Info(msg)
+
+	type Modules struct {
+		Modules []string
+	}
+	mods := Modules{}
+
+	//Modules to insmod during init, paths relative to the kernel folder
+	modules := []string{
+		"/kernel/fs/isofs/isofs.ko",
+		"/kernel/drivers/cdrom/cdrom.ko",
+		"/kernel/drivers/scsi/sr_mod.ko",
+		"/kernel/fs/overlayfs/overlay.ko",
+	}
+
+	/* Find kernel, then break the name into kernelVersion */
+	kernelGlob, err := filepath.Glob(tmpPaths[clrRootfs] + "/lib/kernel/org.clearlinux.*")
+	if err != nil || len(kernelGlob) != 1 {
+		prg.Failure()
+		log.Error("Failed to determine kernel revision or > 1 kernel found")
+		return err
+	}
+	kernelTypeVersion := strings.SplitAfter(filepath.Base((kernelGlob[0])), "org.clearlinux.")[1]
+	kernelType := strings.Split(kernelTypeVersion, ".")[0] //kernelType examples: native,kvm,lts2018,hyperv
+	kernelVersion := strings.SplitAfter(kernelTypeVersion, kernelType+".")[1]
+
+	/* Copy files to initrd, and add to mods so they're added to the init template */
+	for _, i := range modules {
+		rootfsModPath := tmpPaths[clrRootfs] + "/usr/lib/modules/" + kernelVersion + "." + kernelType + i
+
+		/* copy kernel module to initramfs */
+		initrdModPath := filepath.Dir(tmpPaths[clrInitrd] + "/usr/lib/modules/" + kernelVersion + "." + kernelType + i)
+
+		if _, err := os.Stat(initrdModPath); os.IsNotExist(err) {
+			err = os.MkdirAll(initrdModPath, os.ModePerm)
+			if err != nil {
+				prg.Failure()
+				return err
+			}
+		}
+
+		err = utils.CopyFile(rootfsModPath, initrdModPath+"/"+filepath.Base(i))
+		if err != nil {
+			prg.Failure()
+			return err
+		}
+		mods.Modules = append(mods.Modules, "/usr/lib/modules/"+kernelVersion+"."+kernelType+i)
+	}
+
+	tmpl, err := ioutil.ReadFile(templatePath + "/initrd_init_template")
+	if err != nil {
+		prg.Failure()
+		log.Error("Failed to open: initrd template at %s\n", templatePath+"initrd_init_template")
+		return err
+	}
+
+	t := template.New("Modules template")
+	t, err = t.Parse(string(tmpl))
+	if err != nil {
+		prg.Failure()
+		log.Error("Failed to parse init's template")
+		return err
+	}
+
+	f, err := os.Create(tmpPaths[clrInitrd] + "/init")
+	if err != nil {
+		prg.Failure()
+		log.Error("Failed to create init file for initrd!")
+		return err
+	}
+	defer func() {
+		_ = f.Close()
+	}()
+
+	err = t.Execute(f, mods)
+	if err != nil {
+		prg.Failure()
+		log.Error("Failed to execute template filling")
+		return err
+	}
+
+	/* Set correct owner and permissions on initrd's init */
+	if err = os.Chown(tmpPaths[clrInitrd]+"/init", 0, 0); err != nil {
+		prg.Failure()
+		return err
+	}
+	if err = os.Chmod(tmpPaths[clrInitrd]+"/init", 0700); err != nil {
+		prg.Failure()
+		return err
+	}
+
+	prg.Success()
+	return err
+}
+
+/* Build initrd image, and copy to the correct location */
+func buildInitrdImage() error {
+	msg := "Building initrd image"
+	prg := progress.NewLoop(msg)
+	log.Info(msg)
+
+	// Determine current user's path so we can revert to it when this function ends
+	currPath, err := os.Getwd()
+	if err != nil {
+		prg.Failure()
+		return err
+	}
+
+	/* find all files in the initrd path, create the initrd */
+	/* The find command must return filenames without a path (eg, must be run in the current dir) */
+	err = os.Chdir(tmpPaths[clrInitrd])
+	if err != nil {
+		prg.Failure()
+		return err
+	}
+
+	args := "sudo find .| cpio -o -H newc | gzip >" + tmpPaths[clrCdroot] + "/images/initrd.gz"
+	_, err = exec.Command("bash", "-c", args).CombinedOutput()
+	if err != nil {
+		prg.Failure()
+		return err
+	}
+
+	err = os.Chdir(currPath)
+	if err != nil {
+		prg.Failure()
+		return err
+	}
+
+	prg.Success()
+	return err
+}
+
+func mkEfiBoot() error {
+	msg := "Building efiboot image"
+	prg := progress.NewLoop(msg)
+	log.Info(msg)
+
+	cmds := [][]string{
+		{"fallocate", "-l", "100M", tmpPaths[clrCdroot] + "/EFI/efiboot.img"},
+		{"mkfs.fat", "-n", "\"CLEAR_EFI\"", tmpPaths[clrCdroot] + "/EFI/efiboot.img"},
+		{"mount", "-t", "vfat", "-o", "loop", tmpPaths[clrCdroot] + "/EFI/efiboot.img", tmpPaths[clrEfi]},
+		{"cp", "-pr", tmpPaths[clrImgEfi] + "/.", tmpPaths[clrEfi]},
+	}
+
+	for _, i := range cmds {
+		err := cmd.RunAndLog(i...)
+		if err != nil {
+			prg.Failure()
+			return err
+		}
+	}
+
+	/* create dirs in new efi img */
+	if _, err := os.Stat(tmpPaths[clrEfi] + "/EFI/systemd"); os.IsNotExist(err) {
+		err = os.MkdirAll(tmpPaths[clrEfi]+"/EFI/systemd", os.ModePerm)
+		if err != nil {
+			prg.Failure()
+			return err
+		}
+	}
+
+	/* Modify loader/entries/Clear-linux-*, add initrd= line and remove ROOT= and rootwait from kernel command line options */
+	entriesGlob, err := filepath.Glob(tmpPaths[clrEfi] + "/loader/entries/Clear-linux-*")
+	if err != nil || len(entriesGlob) != 1 {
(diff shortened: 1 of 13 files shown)


swupd/swupd.go
```diff
@@ -158,11 +158,12 @@ func (s *SoftwareUpdater) Verify(version string, mirror string) error {
 		}
 	}
 
-	err = cmd.RunAndLog(args...)
-	if err != nil {
-		return errors.Wrap(err)
+	if !verifyOnly {
```
mdhorn: Maybe move this up to ~136 and just return nil if verify only

mdhorn:
> To support a menu on Legacy systems, we need the two files libutil.c32 and menu.c32. These files do not exist in Clear Linux at all. The files are currently from a package in Ubuntu. What should we do?

These files can be generated as part of the existing syslinux package we build for Clear Linux. Currently, we only package a small subset of the syslinux binaries which are needed for Clear. The spec file will need to be updated to also package these files and install them in` /usr/share/syslinux`.

isoutils/isoutils.go
```diff
@@ -0,0 +1,517 @@
...
+	log.Info(msg)
+
+	/* TODO: This takes a long time to run, it'd be nice to see it's output as it's running */
+	args := []string{
+		"mksquashfs",
+		tmpPaths[clr_rootfs],
+		tmpPaths[clr_cdroot] + "/images/rootfs.img",
+		"-comp",
+		"lzo",
+		"-wildcards",
```
mdhorn: We can drop wildcards as we do not need them when only excluding top-level directories.

isoutils/isoutils.go
```diff
@@ -0,0 +1,517 @@
...
+		"-ef",
```
mdhorn: Instead of an exclude file, just use
`-e boot/ -e proc/ -e sys/ -e dev/`

This will allow dropping the extra file for the command; and we will not need wildcards.

isoutils/exclude
```diff
@@ -0,0 +1,4 @@
+boot/*
```
mdhorn: This file can be removed if we switch to listing each exclude on the command line; see comment in [isoutils/isoutils.go](url)

isoutils/isoutils.go
```diff
@@ -0,0 +1,517 @@
...
+	//Modules to insmod during init, paths relative to the kernel folder
+	modules := []string{
+		"/kernel/fs/isofs/isofs.ko",
+		"/kernel/drivers/cdrom/cdrom.ko",
+		"/kernel/drivers/scsi/sr_mod.ko",
+		"/kernel/fs/overlayfs/overlay.ko",
+	}
+
+	/* Find kernel, then break the name into kernelVersion */
+	kernelGlob, err := filepath.Glob(tmpPaths[clr_rootfs] + "/lib/kernel/org.clearlinux.native.*")
```
mdhorn: We will need to update this somewhat because we might not always have the native kernel; could be kvm or lts or ...:
```
# ls /lib/kernel/org.clearlinux.*
/lib/kernel/org.clearlinux.kvm.5.0.3-315
/lib/kernel/org.clearlinux.lts2018.4.19.30-15
/lib/kernel/org.clearlinux.native.5.0.3-721
```

isoutils/isoutils.go
```diff
@@ -0,0 +1,517 @@
...
+	defer f.Close()
+
+	err = t.Execute(f, mods)
+	if err != nil {
+		prg.Failure()
+		log.Error("Failed to execute template filling")
+		return err
+	}
+
+	/* Set correct owner and permissions on initrd's init */
```
mdhorn: We should make these native go calls with os.Chmod and os.Chown

isoutils/isoutils.go
```diff
@@ -0,0 +1,517 @@
...
+	prg := progress.NewLoop(msg)
+	log.Info(msg)
+
+	type BootConf struct {
+		Options string
+	}
+	bc := BootConf{}
+
+	/* Find kernel path so we can copy the kernel later */
+	kernelGlob, err := filepath.Glob(tmpPaths[clr_rootfs] + "/lib/kernel/org.clearlinux.native.*")
```
mdhorn: We will need to update this somewhat because we might not always have the native kernel; could be kvm or lts or ...:
```
# ls /lib/kernel/org.clearlinux.*
/lib/kernel/org.clearlinux.kvm.5.0.3-315
/lib/kernel/org.clearlinux.lts2018.4.19.30-15
/lib/kernel/org.clearlinux.native.5.0.3-721
```

isoutils/boot.txt
```diff
@@ -0,0 +1,3 @@
+
```
mdhorn: Given the relatively small size, I would prefer we just write this file with a native go API instead of storing and copying this file.

mdhorn:
> I could probably move more things from RunAndLog to Go native - cp and umount for example. Suggestions?

Yes, especially the `cp` commands can be replaced with our native `utils.CopyFile()` command.

mdhorn:
> What do I do with files that are needed for execution and need to be available on the filesystem?
> 
> Examples are: isoutils/exclude, isoutils/initrd_init_template, isoutils/isolinux.cfg.template, and both *.c32 files

The remaining template files should be installed to /usr/share/clr-installer/iso
We currently have a themes dirertory which gets installed into /usr/share/clr-installer/themes
We should probably refactor and have a `usr` directory in the source similar to our `etc`, and place both a `iso` as well as the `themes` dirctories under that directory.
You can borrow the existing `utils.lookupThemeDir()` to create a 'utils.lookupISODir`.

mdhorn:
> There are many image types that cannot have an ISO made from them at the moment, because there's a hard requirement on the image supporting EFI and the EFI partition must be mounted at '/boot' in relation to the new image's filesystem base. Any suggestions on how to report to the user/deal with this better?

In this case I would not support images which have the LegacyBios flag set to true.

mdhorn:
> There's now more prerequisites for clr-installer (but only if you intend to make ISO files). These programs are mksquashfs, cpio, mkfs.fat, xorriso and the bundle without executables is bootloader(for isolinux files). This list may not be complete.

We need to test this on an image witih minimum bundles installed to ensure we have the requirements met. Anything missing will need to be added to the clr-installer bundle.
Looks like we will need to add:
```
squashfs-tools-bin
cpio-bin
xorriso-bin
```
We already have mkfs.fat as we use this to create /boot.

mdhorn:
> The size of the ramfs (after boot) is currently hardcoded in the initrd_init_template (line 139), should this be a config option in the YAML/somewhere? Is 256M enough?

This could a reasonable item to add in the YAML config and the command line while still defaulting to the current value of 256MB.
maybe `--iso-ram` for the command line and `isoRam` in the YAML 
Please to update the `scripts/InstallerYAMLSyntax.md` file with the new value.

mdhorn:
@evil-brain This is awesome so far, thanks for the contributions. Let me know if you need clarifications on my comments.

swupd/swupd.go
evil-brain: Done.

isoutils/isoutils.go
evil-brain: Done.

isoutils/isoutils.go
evil-brain: Done - Originally I had done it the other way to preserve the directory entries.

It appears that Clear's init system is "smart" enough to make these directories as required, the system boots and nothing is unusual about the mount points.

isoutils/exclude
evil-brain: Done

isoutils/isoutils.go
evil-brain: Fixed, this is now generic and doesn't rely on native - I tested with kvm.yaml and can produce an ISO now.

isoutils/isoutils.go
evil-brain: Done

isoutils/isoutils.go
evil-brain: Done, same as previous - works with any kernel names now (as long as there's not an extra `.` in them...)

isoutils/boot.txt
evil-brain: Done.

evil-brain:
> Yes, especially the `cp` commands can be replaced with our native `utils.CopyFile()` command.

There's about 10 cp commands, currently in progress.

>You can borrow the existing utils.lookupThemeDir() to create a 'utils.lookupISODir`.
>
>NOTE: We should consider having this rebased against dev-gui branch as it has newer code and is currently our active development work for the new release of the installer.

Will rebase on dev-gui and implement something like utils.LookupThemeDir().

> In this case I would not support images which have the LegacyBios flag set to true.

I can include some kind of gate to disallow LegacyBios enabled images.

>We need to test this on an image witih minimum bundles installed to ensure we have the requirements met.

I'll do this, I already have qemu-kvm configured an in use, it'll be really easy from this point. I'll make note of the required files.

>This could a reasonable item to add in the YAML config and the command line while still defaulting to the current value of 256MB.
>maybe --iso-ram for the command line and isoRam in the YAML
>Please to update the scripts/InstallerYAMLSyntax.md file with the new value.

I'll also look into this.

Current plan is:
remove cp commands -> gate on LegacyBios ->  add iso-ram and associated YAML change -> Rebase and submit -> Test minimal system.

If there's a priority modification, please correct me!

evil-brain:
New push! 
Now rebased on dev-gui!

I addressed all issues above except for:
 * Adding iso-ram option
 * create a 'utils.lookupISODir` function to handle the template directory.

evil-brain:
Added utils.LookupISOTemplateDir() to handle the template file paths.

isoutils/isoutils.go
```diff
@@ -0,0 +1,539 @@
...
+	return err
+}
+
+func mkEfiBoot() error {
+	msg := "Building efiboot image"
+	prg := progress.NewLoop(msg)
+	log.Info(msg)
+
+	cmds := [][]string{
+		[]string{"fallocate", "-l", "800M", tmpPaths[clr_cdroot] + "/EFI/efiboot.img"},
```
evil-brain: This really doesn't need to be 800M, I think I got this size from the previous ISO script.

Currently the size of the *FILES* installed to this image is 32MB, I'd suggest probably 100MB?
Or maybe there's some way to resize this FS to the content size after creation?

The current ISO for a live image is 2.7GB. When xz'd the size is 1.5GB. Most of this compression is compressing this huge, empty EFI boot partition.

evil-brain:
Another force-push, updated the ramfs from 256MB to 512MB - more room to do activities and still fairly low ram usage.
I doubt many will boot this on a system with <4GB (because 2GB will be tempfs, .5G is ramfs root, which leaves 1.5G for system RAM)

mdhorn:
> Another force-push, updated the ramfs from 256MB to 512MB - more room to do activities and still fairly low ram usage.
> I doubt many will boot this on a system with <4GB (because 2GB will be tempfs, .5G is ramfs root, which leaves 1.5G for system RAM)

@evil-brain Should we consider adding a memory size check with the other Clear Linux OS requirements in the init? We might also want to internally verify that 4GB minimum for a desktop or server is acceptable.

evil-brain:
> > Another force-push, updated the ramfs from 256MB to 512MB - more room to do activities and still fairly low ram usage.
> > I doubt many will boot this on a system with <4GB (because 2GB will be tempfs, .5G is ramfs root, which leaves 1.5G for system RAM)
> 
> @evil-brain Should we consider adding a memory size check with the other Clear Linux OS requirements in the init? We might also want to internally verify that 4GB minimum for a desktop or server is acceptable.

I think we could add a size check - what size should we check against?
Some quick testing on virtualbox with a 512MB ramfs - launching firefox:
512MB: Unusable, processes are being killed due to low memory.
768MB: Boots to desktop, terminal launches, firefox fails to run acceptably
1024MB: Same as 768
1536MB: Firefox works, this would probably be the practical minimum or close to it.

The recommendation on clearlinux.org is 4GB: https://clearlinux.org/documentation/clear-linux/reference/system-requirements

Fail if less than 4GB of RAM, or somehow warn and boot anyways?

mdhorn:
> > > Another force-push, updated the ramfs from 256MB to 512MB - more room to do activities and still fairly low ram usage.
> > > I doubt many will boot this on a system with <4GB (because 2GB will be tempfs, .5G is ramfs root, which leaves 1.5G for system RAM)
> > 
> > 
> > @evil-brain Should we consider adding a memory size check with the other Clear Linux OS requirements in the init? We might also want to internally verify that 4GB minimum for a desktop or server is acceptable.
> 
> I think we could add a size check - what size should we check against?
> Some quick testing on virtualbox with a 512MB ramfs on desktop-beta - launching firefox:
> 512MB: Unusable, processes are being killed due to low memory.
> 768MB: Boots to desktop, terminal launches, firefox fails to run acceptably
> 1024MB: Same as 768
> 1536MB: Firefox works, this would probably be the practical minimum or close to it.
> 
> The recommendation on clearlinux.org is 4GB: https://clearlinux.org/documentation/clear-linux/reference/system-requirements
> 
> Fail if less than 4GB of RAM, or somehow warn and boot anyways?
> It would also be a different value for just the install image for example - I imagine that would run fine on 768.

I'd propose:
- Failing with error when less than 2GB of RAM; make that the hard lower limit.
- Present a warning below the 4GB, but allow them to boot

If we are not already showing that URL of system requirements if they fail the CPU requirements, then we probably should. We should also show it for less than 4GB of memory.

Reasonable?

evil-brain:
> I'd propose:
> 
>     * Failing with error when less than 2GB of RAM; make that the hard lower limit.
> 
>     * Present a warning below the 4GB, but allow them to boot
> 
> 
> If we are not already showing that URL of system requirements if they fail the CPU requirements, then we probably should. We should also show it for less than 4GB of memory.
> 
> Reasonable?

Sounds reasonable to me.

Implementation details are getting a bit difficult - so far I'm using dmidecode (which bloats initrd because it requires the entire 100MB bundle 'sysadmin-basic'), but it's got problems with VMs which don't report memory size via SMBIOS: my command just a returns null result (test with virtualbox). This may also be fragile depending on EFI vs legacy boot.

I could determine memory size by checking `/proc/meminfo`, but it's size is reported as "Total usable RAM (i.e., physical RAM minus a few reserved bits and the kernel binary code)." (man 5 proc)

But this makes it hard to determine values to check against. I may use this method and just do a fuzzy check (ex if value is greater than 2000MB, and hopefully the reserved bits are <48MB or this fails...and hopefully the reserved bits don't ever grow to >48MB)

I added the requirements URL to the failure message, but I am a bit worried one day that URL may change and no one will remember to update it in initrd's init.

mdhorn:
> > I'd propose:
> > ```
> > * Failing with error when less than 2GB of RAM; make that the hard lower limit.
> > 
> > * Present a warning below the 4GB, but allow them to boot
> > ```
> > If we are not already showing that URL of system requirements if they fail the CPU requirements, then we probably should. We should also show it for less than 4GB of memory.
> > Reasonable?
> 
> Sounds reasonable to me.
> 
> Implementation details are getting a bit difficult - so far I'm using dmidecode (which bloats initrd because it requires the entire 100MB bundle 'sysadmin-basic'), but it's got problems with VMs which don't report memory size via SMBIOS: my command just a returns null result (test with virtualbox). This may also be fragile depending on EFI vs legacy boot.
> 
> I could determine memory size by checking `/proc/meminfo`, but it's size is reported as "Total usable RAM (i.e., physical RAM minus a few reserved bits and the kernel binary code)." (man 5 proc)
> 
> But this makes it hard to determine values to check against. I may use this method and just do a fuzzy check (ex if value is greater than 2000MB, and hopefully the reserved bits are <48MB or this fails...and hopefully the reserved bits don't ever grow to >48MB)
> 
> This also may all fail depending on how much RAM is reserved for the iGPU - I'll have to investigate.
> 
> I added the requirements URL to the failure message, but I am a bit worried one day that URL may change and no one will remember to update it in initrd's init.

I'd go with the /proc/meminfo to save on image size.
Just make it fuzzy with a good margin for error.
My testing with a 2G vm shows only 2032348kB for MemTotal and the reserved amount seem to grow slightly more than linear as you increase RAM size.

The URsL are supposed to always have a redirect if there retire a currently published one; we do the best we can.

evil-brain:
If I make this change it will effect all ISO images the same way, but I think it doesn't make much sense to limit the TUI installer on RAM.

A use case would be a user who wants to install os-core and a kernel on a system with 1GB of RAM, the TUI installer and the very basic installed system will work. It seems wrong to block them on an artificial check.

Ubuntu's live image does no RAM checks, but it does silently fail if there's too little system RAM.

I'll work on changing over to checking /proc/meminfo and perhaps this should be submitted as an issue to resolve later (maybe a DisableMemcheck yaml option, or configurable memcheck paramaters)

mdhorn:
> If I make this change it will effect all ISO images the same way, but I think it doesn't make much sense to limit the TUI installer on RAM.
> 
> A use case would be a user who wants to install os-core and a kernel on a system with 1GB of RAM, the TUI installer and the very basic installed system will work. It seems wrong to block them on an artificial check.
> 
> Ubuntu's live image does no RAM checks, but it does silently fail if there's too little system RAM.
> 
> I'll work on changing over to checking /proc/meminfo and perhaps this should be submitted as an issue to resolve later (maybe a DisableMemcheck yaml option, or configurable memcheck paramaters)

Okay, sounds best to drop the memory check for now and maybe at it as a feature enhancement issue to GitHub instead.

Given we drop this memory issue, are you ready change this from a draft PR to final review?
Would like to get this merged soon.

mdhorn:
Please make a copy of the scripts/live-desktop-beta.yaml called scripts/live-desktop.yaml and add the following bundles:
libreoffice
package-utils
pidgin
Remmina
c-basic
gimp

Also add the 'iso' set to 'true' in the new YAML file.
We will use this as the new test target for desktop and as the starting point for the new ISO image that DevOps will generate after the new GUI release.

evil-brain:
> Please make a copy of the scripts/live-desktop-beta.yaml called scripts/live-desktop.yaml and add the following bundles:
> libreoffice
> package-utils
> pidgin
> Remmina
> c-basic
> gimp
> 
> Also add the 'iso' set to 'true' in the new YAML file.
> We will use this as the new test target for desktop and as the starting point for the new ISO image that DevOps will generate after the new GUI release.

I'm looking at this right now, but the image is inexplicably much larger than it used to be. It's so large at this point the 7.8GB size defined in the yaml is too small.

In version 28750 I could build an image with these extra bundles and there was no problem, but in 28790 the rootfs is 9.8GB, previously it was 6.3GB

output from du for the two images (rootfs.img is compressed, squashfs-root is uncompressed):
2.0G    28750/rootfs.img
6.3G    28750/squashfs-root
2.8G    28830/rootfs.img
9.8G    28830/squashfs-root

Investigation:
/usr/lib32 went from 3.0M to 3.5G in this time - some bundle is including 32-bit lib support.
...still investigating, will edit when I have more info...

Bloat seems like a problem, and we're really getting close to the 8GB(minus other filesystems) limit. Should we increase the limit (.img files won't fit on 8GB sticks any more)?

mdhorn:
> > Please make a copy of the scripts/live-desktop-beta.yaml called scripts/live-desktop.yaml and add the following bundles:
> > libreoffice
> > package-utils
> > pidgin
> > Remmina
> > c-basic
> > gimp
> > Also add the 'iso' set to 'true' in the new YAML file.
> > We will use this as the new test target for desktop and as the starting point for the new ISO image that DevOps will generate after the new GUI release.
> 
> I'm looking at this right now, but the image is inexplicably much larger than it used to be. It's so large at this point the 7.8GB size defined in the yaml is too small.
> 
> In version 28750 I could build an image with these extra bundles and there was no problem, but in 28790 the rootfs is 9.8GB, previously it was 6.3GB
> 
> output from du for the two images (rootfs.img is compressed, squashfs-root is uncompressed):
> 2.0G 28750/rootfs.img
> 6.3G 28750/squashfs-root
> 2.8G 28830/rootfs.img
> 9.8G 28830/squashfs-root
> 
> Investigation:
> /usr/lib32 went from 3.0M to 3.5G in this time - some bundle is including 32-bit lib support.
> ...still investigating, will edit when I have more info...
> 
> Bloat seems like a problem, and we're really getting close to the 8GB(minus other filesystems) limit. Should we increase the limit (.img files won't fit on 8GB sticks any more)?

The goal was an 8GB USB drive -- the bundle selections were picked based on this size requirement.

If you do not get a quick identification on the bloat due to 32-bit, please raise this on the mailing list and IRC and someone should be able to help track this down.

evil-brain:
I tracked it down and asked on IRC. It seems this is a mistake and will be fixed in future releases.
<arjan> we're working on getting the 32 bit stuff out of llvm again ;(

No timeline, but I'll still submit the yaml changes.

evil-brain:
Fixed linter issues, retrying. I will babysit Travis until this passes or fails for a reason I can't control or don't understand.

evil-brain:
Fixed linter issues, retrying. I will babysit Travis until this passes or fails for a reason I can't control or don't understand.

mdhorn:
@evil-brain Besides a new rebase against dev-gui to get the fix from #285, we need to add a patch to the code for parsing the command line flags. The original patch I gave you has a logic error, and the raw image file was being removed when not generating an ISO.

Third minor item, the ISO should be generated in the same location as the original raw image. Currently, the ISO is being placed in the same directory with the installer executable; please fix this.
Also, it looks like when we do make an ISO, it is not removing the raw image:
`2019/04/17 23:28:56 [WRN] Failed to remove image file: installer.img`
My guess would be it is assuming the raw image is the same location as the ISO which is currently not true from above. After fixing the location where the ISO lands, verify that the raw image is removed.

PR #275, opened 2019-04-02
gui: Display language as string

mdhorn:
- Display the language as a string like the TUI instead of a code


gui/pages/language.go
```diff
@@ -214,5 +219,6 @@ func (page *LanguagePage) GetConfiguredValue() string {
 	if page.model.Language == nil {
 		return ""
 	}
-	return page.model.Language.Code
+
+	return page.model.Language.String()
```
reaganlo: In order for the summary to look more neat and aligned, can you please replace the return statement with the following: 	
```
whtspExp := regexp.MustCompile(`[\s\p{Zs}]{2,}`) // Additional white space inside string
return whtspExp.ReplaceAllString(page.model.Language.String(), " ")
```
We could have this at a common place https://github.com/clearlinux/clr-installer/blob/dev-gui/gui/summary_widget.go#L145 but then it messes up with the bundle summary and makes them all in the same line.

mdhorn:
An updated version of this patch sent to Reagan to be added to the new localization PR.

Issue #14, opened 2018-08-13
Support dual OS


mdhorn:
<a href="https://github.com/dorileo"><img src="https://avatars.githubusercontent.com/dorileo" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [dorileo](https://github.com/dorileo)**
_Tuesday Apr 10, 2018 at 20:02 GMT_

----

Request to be able to install Clear Linux on a media where an existing OS already resides.

mdhorn:
<a href="https://github.com/mdhorn"><img src="https://avatars.githubusercontent.com/mdhorn" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mdhorn](https://github.com/mdhorn)**
_Friday Jul 20, 2018 at 23:55 GMT_

----


The technical scope of what is possible due to Clear Linux booting requirements need to be reviewed prior to committing to this request.

mdhorn:
The primary goal is to meet similar requirements to what was possible in ister_gui.
- Install Clear Linux OS onto a target media with an existing OS present
-- Only update the GUID Partition Table (GPT) with newly added partition(s)
-- Only create file systems on new partition(s)

Open Items:
- Enable reuse of existing partitions (swap, /home, /srv, ...)

mvincerx:
We're eager to implement this change in the docs too. This would greatly improve multi-boot scenarios, especially with Windows. I have an open PR on this: https://github.com/clearlinux/clear-linux-documentation/pull/382

mdhorn:
The partitioning changes are now supported in the User Interface.
Not sure if anything else can be done for this topic in the installer unless Clear Linux* OS improves or considers supporting other boot loaders.

PR #278, opened 2019-04-09
Partitioning

mdhorn:
Fixes Issue: #14 

- New TUI interface for media selection
- Backend changes to ensure we do not overwrite existing media during Safe Install
- Media will be overwritten if Destructive Install is chosen by the user
- Launch disk partitioning tool for selected media to allow the user to create free space for the install
  - Note: Resizing a partition does NOT resize a file system by default


model/model.go
```diff
@@ -41,6 +41,12 @@ var testAlias = []string{}
 // SystemInstall represents the system install "configuration", the target
 // medias, bundles to install and whatever state a install may require
 type SystemInstall struct {
+	// Even though TargetMedias is an array, we currently only support
+	// it containing a single BlockDevice which matches/matps to the InstallTarget
```
reaganlo: Spell check  "matches/maps"

mdhorn: fixed

tui/disk_util.go
```diff
@@ -0,0 +1,333 @@
...
+	contentFrame.SetPack(clui.Vertical)
+
+	page.descrptionLabel = clui.CreateLabel(contentFrame, AutoSize, 7, "", Fixed)
+	page.descrptionLabel.SetMultiline(true)
+	page.descrptionLabel.SetPaddings(0, 2)
+	descriptiveMessage := "Launch an external program (" + diskUtil +
+		") to modify the selected\n" +
+		"disk in an attempt to allocate enough free disk space to allow\n" +
+		"the installation to succeed.\n\n" +
+		"Only disk with a 'gpt' partition can be modifify.\n\n" +
```
reaganlo: Spellcheck "modify" instead of "modifify"

tui/media_config.go
```diff
@@ -0,0 +1,427 @@
...
+					warning = fmt.Sprintf("Warning: %s", warning)
+					page.labelWarning.SetTitle(warning)
+				}
+			}
+		}
+	})
+
+	// Description of Safe
+	descLabel := clui.CreateLabel(radioButtonFrame, 2, 3, "Uses non-partitioned or gpt partitioned media with free space.\n"+
+		"Install along side existing data or OS if adequate space is available.", Fixed)
```
reaganlo: Usage. Change "along side" to one word "alongside"

tui/disk_util.go
mdhorn: fixed

tui/media_config.go
mdhorn: fixed in new force push

reaganlo:
Few things Mark:
- Is it intentional to have the media list change only on "**double click**" of the Installation Type instead of "**single click**"?
-  **Auto partition** gives a warning and a **Dismiss** button. Not able to proceed further with the Auto partition. Is that intentional?
- During manual partition, when I select each partition and reduce the size (e.g by a few MB's), it gives the message size too large.

mdhorn:
> @mdhorn Few things Mark:
> 
> * Is it intentional to have the media list change only on "**double click**" of the Installation Type instead of "**single click**"?
> * **Auto partition** gives a warning and a **Dismiss** button. Not able to proceed further with the Auto partition. Is that intentional?
> * During manual partition, when I select each partition and reduce the size (e.g by a few MB's), it gives the message size too large.

I can't seem to find the "double click" issue -- I'll stop by.  But in general, the TUI does not officially support mouse, only keyboard.

The Auto Partition warns because you are changing a disk with content. After it is dismissed, the disk you selected should have already been partitioned and you are ready to hit the Confirm button. If that's not the case, I need to see your test case.

The "size too large" is a bug in that we switched everything in manual partitioning from MiB to MB, but the check on the size was still checking MiB. I'm pushing a fix for this.

Issue #232, opened 2019-02-21
Add clear-linux-check-config.sh to test HW compatibility with CL


tpleavitt:
**Is your feature request related to a problem? Please describe.**
Some CPUs do not support specific instruction sets, for example an IVB CPU that does not support AES, even though this CPU falls under the umbrella of supported CPUs.  CL will install but applications will fail that need AES functionality.  See https://github.com/clearlinux/distribution/issues/471 for a specific example.

**Describe the solution you'd like**
Add this check to the installer as an option to run the HW check prior to doing an install

**Describe alternatives you've considered**
Add the shell script that checks HW compatibility to the desktop as an item that can be run by the user before running the TUI installer.

**Additional context**

ahkok:
I'd rather have the installer actually test whether the CPU is compatible by checking for the required instruction sets, and fail if it isn't. Not having AES will lead to very subtle and difficult to debug problems, as you have seen.

mdhorn:
The system check was added to both the ISO as a pre-boot check as well as integrated as a pre-check in both user interfaces.
Additionally, one change run the pre-check manually by using the `--system-check` flag.

Note: The AES requirement was removed from Clear Linux in Late 2019 and removed from the installer with PR #500

PR #288, opened 2019-04-19
Add system compatibility checking

evil-brain:
Check the system meets Clear Linux's requirements before running
install.

Enable the sysCheck in the default clr-installer.yaml as well.

Currently, this checks for required CPU features and for EFI firmware
(via sysfs).

Fixes Issue: #232 

- Add syscheck package, run by default as part of Install.

Current issues:
 - Failure mode is pretty ugly with a Panic message and stack dump
 - Tests only EFI firmware and CPU features. No other checks (RAM), can only fail and not warn.
 - Failure output could be more descriptive, current looks like "Missing CPU feature: aes"

### DIFF
--- a/args/args.go
+++ b/args/args.go
@@ -71,6 +71,7 @@ type Args struct {
 	MakeISOSet              bool
 	KeepImage               bool
 	KeepImageSet            bool
+	SystemCheck             bool
 }
 
 func (args *Args) setKernelArgs() (err error) {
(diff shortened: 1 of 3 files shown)


mdhorn:
@evil-brain We just rebased dev-gui to include the latest code from the master branch.

You will have to rebase.

model/model.go
```diff
@@ -77,6 +77,7 @@ type SystemInstall struct {
 	CryptPass         string                 `yaml:"-"`
 	MakeISO           bool                   `yaml:"iso,omitempty,flow"`
 	KeepImage         bool                   `yaml:"keepImage,omitempty,flow"`
+	SystemCheck       bool                   `yaml:"systemCheck,omitempty,flow"`
```
mdhorn: Maybe we should drop this from the configuration file for now since we only support running this from the command line standard alone?

evil-brain: Done!

PR #291, opened 2019-04-22
gui: Make User Add page as Required

reaganlo:
- Make User Add page as Required

Signed-off-by: Reagan Lopez <reagan.lopez@intel.com>


reaganlo:
@mdhorn Made the change as task not completed if no users are added. Good catch. Regarding the Title name, as discussed, since we also have the option to delete the user, I would suggest we keep it as "Manage User"

PR #302, opened 2019-04-24
gui: Simply password check

mdhorn:
- Use the same code to check the password and confirm password.


mdhorn:
@reaganlo This removes some redundant code but also ensure we verify the password in the same way when either the Password or Confirm is updated.  Bun has commented on this.

PR #310, opened 2019-04-26
swupd: Speed up install

mdhorn:
- Use new bundle option to swupd verify instead of bundle adds.
- Optimize the flow to only have a single swupd verify call.


mdhorn:
I just test the interactive build and it goes much faster, but users will not like the fact it sits for 5-10 minutes on the "Installing base OS and configured bundles".
If we do not get the swupd real-time updates working soon, I might at go add a spinner or something so they do not perceive it is hung.

PR #311, opened 2019-04-26
gui: Add/update localization strings

reaganlo:
- Adds localization for bundle description
- Adds localization for variables
- Removes unsused localization strings
- Adds missing localization strings
- Standardize the strings used for rules
- TODO: Determine if Timezone and Keyboard values should be translated

### DIFF
--- a/gui/pages/hostname.go
+++ b/gui/pages/hostname.go
@@ -47,7 +47,7 @@ func NewHostnamePage(controller Controller, model *model.SystemInstall) (Page, e
 	page.box.PackStart(page.entry, false, false, 0)
 
 	// Rules label
-	rulesText := utils.Locale.Get("You can use alpha-numeric characters and hyphens with a maximum of 63 characters")
+	rulesText := utils.Locale.Get("Can use alpha-numeric characters and - with a maximum of %d characters.", hostname.MaxHostnameLength)
 	page.rules, err = setLabel(rulesText, "label-rules", 0.0)
 	if err != nil {
 		return nil, err
--- a/gui/pages/user_add.go
+++ b/gui/pages/user_add.go
@@ -76,7 +76,7 @@ func NewUserAddPage(controller Controller, model *model.SystemInstall) (Page, er
 
 	// Name
 	page.name, page.nameWarning, err = page.setSimilarWidgets(utils.Locale.Get("User Name"),
-		utils.Locale.Get("Must start with letter. Can use numbers, hyphens and underscores. Max %d characters.", user.MaxUsernameLength),
+		utils.Locale.Get("Must start with letter. Can use numbers, commas, - and _. Max %d characters.", user.MaxUsernameLength),
 		user.MaxUsernameLength)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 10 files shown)


gui/pages/hostname.go
```diff
@@ -47,7 +47,7 @@ func NewHostnamePage(controller Controller, model *model.SystemInstall) (Page, e
 	page.box.PackStart(page.entry, false, false, 0)
 
 	// Rules label
-	rulesText := utils.Locale.Get("You can use alpha-numeric characters and hyphens with a maximum of 63 characters")
+	rulesText := utils.Locale.Get("Can use alpha-numeric characters and - with a maximum of %d characters.", hostname.MaxHostnameLength)
```
mdhorn: Without the "You" this seems strange.
What about:
"Valid characters are alpha-numerics and hyphens '-' with a maximum of %d characters."
or even:
"Valid characters are alpha-numerics and '-' with a maximum of %d characters."

mdhorn: Or if for consistency we leave the "Can use", we should at least single quote any characters we reference like:
 "Can use alpha-numeric characters and '-' with a maximum of %d characters."

gui/pages/user_add.go
```diff
@@ -76,15 +76,15 @@ func NewUserAddPage(controller Controller, model *model.SystemInstall) (Page, er
 
 	// Name
 	page.name, page.nameWarning, err = page.setSimilarWidgets(utils.Locale.Get("User Name"),
-		utils.Locale.Get("Must start with letter. Can use numbers, hyphens and underscores. Max %d characters.", user.MaxUsernameLength),
+		utils.Locale.Get("Must start with letter. Can use numbers, commas, - and _. Max %d characters.", user.MaxUsernameLength),
```
mdhorn: So I didn't add the 'commas' originally because that is a field separator and didn't care to explain the nuances. If you are going to add it, it should be a character like the others.
We should again single quote all of the valid characters for clarity.
"Must start with letter. Can use numbers, ',', '-' and '_'. Max %d characters."

gui/pages/user_add.go
```diff
@@ -76,15 +76,15 @@ func NewUserAddPage(controller Controller, model *model.SystemInstall) (Page, er
...
 		user.MaxUsernameLength)
 	if err != nil {
 		return nil, err
 	}
 
 	// Login
 	page.login, page.loginWarning, err = page.setSimilarWidgets(utils.Locale.Get("Login")+" *",
-		utils.Locale.Get("Must start with letter. Can use numbers, hyphens and underscores. Max %d characters.", user.MaxLoginLength),
+		utils.Locale.Get("Must start with letter. Can use numbers, - and _. Max %d characters.", user.MaxLoginLength),
```
mdhorn: Same here:
"Must start with letter. Can use numbers, '-' and '_'. Max %d characters."

gui/pages/user_add.go
reaganlo: @mdhorn I was trying to keep the rules msg similar to the warning message returned by the validation functions. E.g in IsValidUsername(). So should I be changing it everywhere to have single quotes?

gui/pages/hostname.go
reaganlo: @mhorn Or maybe like the warning msg I can explicitly mention the field name.. E.g "Hostname can contain..."

gui/pages/user_add.go
mdhorn: @reaganlo hmmm... maybe these rule strings belong in the non-UI class?
Or can they be constructed using the valid calls on known bad values to get the strings?
rule := IsValidUsername("") + IsValidUsername("super long string that is invalude")?

mdhorn: After our discussion, I'm okay with what we have already in the pull request.

PR #322, opened 2019-04-28
gui: Desktop Launcher fixes

mdhorn:
- Made all files and actions consistent with the clr-installer-gui name.
- Fix security of starting Installer; user pkexec not sudo; added new rule for clr-installer-gui.
- Fix bug so only one icon is present when running.

### DIFF
--- a/etc/org.clearlinux.clr-installer-gui.policy
+++ b/etc/org.clearlinux.clr-installer-gui.policy
@@ -3,12 +3,16 @@
 "http://www.freedesktop.org/software/polkit/policyconfig-1.dtd">
 <policyconfig>
 
-  <vendor>Clear Linux OS Installer</vendor>
+  <vendor>Clear Linux* OS Installer</vendor>
   <vendor_url>https://github.com/clearlinux/clr-installer</vendor_url>
 
-  <action id="org.clearlinux.clr-installer.start">
-    <description>Install Clear Linux OS</description>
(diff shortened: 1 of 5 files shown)


etc/org.clearlinux.clr-installer-gui.policy
```diff
@@ -3,12 +3,16 @@
...
-  <vendor>Clear Linux OS Installer</vendor>
+  <vendor>Clear Linux* OS Installer</vendor>
   <vendor_url>https://github.com/clearlinux/clr-installer</vendor_url>
 
-  <action id="org.clearlinux.clr-installer.start">
-    <description>Install Clear Linux OS</description>
-    <message>Authentication is required to install Clear Linux OS</message>
+  <action id="org.clearlinux.clr-installer-gui.start">
+    <description>Install Clear Linux* OS</description>
+    <description xml:lang="es">MX: Install Clear Linux* OS</description>
```
reaganlo: Shouldn't we use "es_MX" instead of just "es"

mdhorn: I can change this to the specific es_MX.
The reason I had left it as just es was that the other example present only has 'es'.
If you use the Desktop Settings application, and select Language the Spanish one shown by default is just 'es';  you can click the ... to search for Spanish Mexico.
I'll change it for consistency.

PR #324, opened 2019-04-29
isoutils: Include custom kernel command line parms

evil-brain:
isoutils had a bug which would remove additional kernel command line
paramaters specified during image build time.

To fix this, instead of reading the cmdline file from the rootfs,
read the modified cmdline file from the EFI partition (which
clr-boot-manager fills in for us).

Modify the options as required, and then write in the appropriate
places.

- Read the EFI partition from the already-created image, modify as needed, use for new boot systems.

### DIFF
--- a/isoutils/isoutils.go
+++ b/isoutils/isoutils.go
@@ -302,7 +302,7 @@ func mkEfiBoot() error {
 		}
 	}
 
-	/* Modify loader/entries/Clear-linux-*, add initrd= line and remove ROOT= and rootwait from kernel command line options */
+	/* Modify loader/entries/Clear-linux-*, add initrd= line and remove ROOT= from kernel command line options */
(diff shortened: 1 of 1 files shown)


isoutils/isoutils.go
```diff
@@ -434,7 +431,24 @@ func mkLegacyBoot(templatePath string) error {
...
+	var optionsLine string
+	for _, line := range lines {
+		if strings.Contains(line, "options") {
+			optionsLine = line
+		}
+	}
+
+	options := strings.Split(optionsLine, " ")
+	for i, option := range options {
+		if strings.Contains(option, "options") {
```
evil-brain: Doesn’t this still have root=PARTUUID option?

evil-brain:
Fixed some issues, I was confused over mkEfiBoot reading/writing a file and mkLegacyBoot reading a file but filling a template.

I've re-tested this (booting both EFI and Legacy and booting the img in EFI), and this is functional.

PR #339, opened 2019-05-02
gui: Add second set of translations

reaganlo:
gui: Add second set of translations


mdhorn:
@reaganlo Please rebase with the new string which was missing "Disabled" against #340

PR #343, opened 2019-05-02
Fix a bug in Partial truncating an integer value

castulo:
The Partial() function is used to report a partial progress in a
progress bar. When using this function the value set on the progress bar
was always zero since it was doing a division of two integers that were
always resulting in a number < 1 which was being truncated because both
numbers were integers so go was using an integer to store the result
too.

This commit fixes the bug by converting the values to the right types.


mdhorn:
Good catch!

PR #344, opened 2019-05-02
Use the JSON output from swupd to report progress

castulo:
When running the "swupd verify" command to install the base OS and the
selected bundles, the progress of the command execution was just being
shown as a progress bar that was looping until the swupd command
finished. This approach was not ideal since this process can take
several minutes to complete, and since the progress bar was just
spinning it didn't provide a real feedback for the user to know in what
part the process was. This can trigger some anxiety in users.

This commit solves this problem by using the machine readable output
from swupd to determine the exact part of the process where the install is
and report this to the user using a series of steps along with a progress
bar (or percentage value in not TUI installer) that shows how far the
step has gotten.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -5,6 +5,7 @@
 package cmd
 
 import (
+	"bufio"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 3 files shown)


mdhorn:
@castulo This will need to be rebased on 'master' branch as we've EOL'd dev-gui today.

castulo:
I will close this PR and will submit a new one against master.

PR #345, opened 2019-05-02
isoutils: Add Rufus ISO image support

evil-brain:
Rufus in iso image mode creates a new filesystem and copies all files
from the iso onto that filesystem. This broke Clear boot because there
were missing files from the CD root to boot (those files were in
efiboot.img), and because initrd was limited to mounting only ISO files
to find the rootfs.

Now the files from efiboot.img are copied to the ISO root (so that Rufus
will now work), the initrd paths have been made congruent (though there
are two initrd.gz files - one in efiboot.img and one on the ISO root),
and some extraneous systemd files have been removed.

- Fix Rufus' ISO Image mode
- Make initrd paths on kernel cmdline congruent
- Remove extraneous systemd boot files

This may have a side-effect of bloating the image slightly (by probably 32MB), but I can't see a way to avoid this as having two copies is crucial to support all 4 boot methods (Rufus ISO/dd) and (efi/legacy)

### DIFF
--- a/iso_templates/isolinux.cfg.template
+++ b/iso_templates/isolinux.cfg.template
@@ -4,7 +4,7 @@ TIMEOUT 50
 
 LABEL clear
   MENU DEFAULT
-  MENU LABEL Clear Linux OS for Intel Architecture
+  MENU LABEL Clear Linux* OS
   LINUX /kernel/kernel.xz
-  INITRD /images/initrd.gz
+  INITRD /EFI/BOOT/initrd.gz
   APPEND {{.Options}}
--- a/isoutils/isoutils.go
+++ b/isoutils/isoutils.go
@@ -251,7 +251,16 @@ func buildInitrdImage() error {
 		return err
 	}
 
-	args := "sudo find .| cpio -o -H newc | gzip >" + tmpPaths[clrCdroot] + "/images/initrd.gz"
+	initrdPath := tmpPaths[clrCdroot] + "/EFI/BOOT/"
+	if _, err := os.Stat(initrdPath); os.IsNotExist(err) {
+		err = os.MkdirAll(initrdPath, os.ModePerm)
+		if err != nil {
+			prg.Failure()
+			return err
+		}
+	}
+
+	args := "sudo find .| cpio -o -H newc | gzip >" + initrdPath + "initrd.gz"
 	_, err = exec.Command("bash", "-c", args).CombinedOutput()
 	if err != nil {
 		prg.Failure()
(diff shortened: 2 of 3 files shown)


isoutils/isoutils.go
```diff
@@ -256,7 +256,16 @@ func buildInitrdImage() error {
...
+	initrdPath := tmpPaths[clrCdroot] + "/EFI/BOOT/"
+	if _, err := os.Stat(initrdPath); os.IsNotExist(err) {
+		err = os.MkdirAll(initrdPath, os.ModePerm)
+		if err != nil {
+			prg.Failure()
+			return err
+		}
+	}
+
+	args := "sudo find .| cpio -o -H newc | gzip >" + tmpPaths[clrCdroot] + "/EFI/BOOT/initrd.gz"
```
evil-brain: Change this to use initrdPath, since I defined it earlier...

isoutils/isoutils.go
```diff
@@ -343,20 +343,27 @@ func mkEfiBoot() error {
 		return err
 	}
 
-	/* Copy all required files to efiboot.img and finally unmount efiboot.img */
-	paths := [][]string{
-		{tmpPaths[clrCdroot] + "/images/initrd.gz", tmpPaths[clrEfi] + "/EFI/Boot/initrd.gz"},
-		{tmpPaths[clrImgEfi] + "/EFI/BOOT/BOOTX64.EFI", tmpPaths[clrEfi] + "/EFI/systemd/systemd-bootx64.efi"},
+	/* Copy EFI files to the cdroot for Rufus support */
+	cmds = [][]string{
```
evil-brain: This doesn't need to be a slice of slices

mdhorn:
@evil-brain This will need to be rebased on 'master' branch as we've EOL'd dev-gui today.

evil-brain:
> @evil-brain This will need to be rebased on 'master' branch as we've EOL'd dev-gui today.

Done, other fixes I wanted to make are done as well.

The only thing that could be better about this is figuring out Xorriso's -e option so I can eliminate the entire EFI image file. That would save us about 100MB (32MB if they EFI file were resized after creation)

iso_templates/isolinux.cfg.template
```diff
@@ -6,5 +6,5 @@ LABEL clear
   MENU DEFAULT
   MENU LABEL Clear Linux OS for Intel Architecture
```
mdhorn: Please change to:
"Clear Linux* OS"

isoutils/isoutils.go
mdhorn: change to a single array of commands and remove the for loop as there is only one command now.

iso_templates/isolinux.cfg.template
evil-brain: Done, updated all other instances I could find as well.

PR #348, opened 2019-05-03
gui: Removes lag due to scanning media

reaganlo:
- Scans for disk during installer launch
- Also ensures that encryption is clickable only within its label


reaganlo:
Closing this PR as it requires more testing and work.

PR #352, opened 2019-05-04
user: Only allow lower case letter for login

mdhorn:
Recently discovered Clear Linux only allows lower case letter
as part of a login name. Updated validation check and error message.


mdhorn:
> Should we update the rules too?

I was avoiding doing two new translation strings.  The rule is still valid:
"Must start with letter. Can use numbers, - and _. Max %d characters."
it just isn't as specific as the error message.
If you think we should add "lowercase" to the rule, should we also adjust the grammar?
"Must start with a lowercase letter. Can use numbers, - and _. Max %d characters."

reaganlo:
How did you translate the first one? :)

mdhorn:
> How did you translate the first one? :)
Google. 🎲

PR #361, opened 2019-05-07
iso: Pass all options through to ISO generation

mdhorn:
We need to pass the full suite of options to the ISO
generation code to ensure the call to swupd has all
of the requested options.


mdhorn:
Approved for push -- validated in DevOps

Issue #12, opened 2018-08-13
implement swupd real progress


mdhorn:
<a href="https://github.com/dorileo"><img src="https://avatars.githubusercontent.com/dorileo" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [dorileo](https://github.com/dorileo)**
_Wednesday Mar 28, 2018 at 19:20 GMT_

----

Now we do a "loop based" progress for steps involving swupd commands, we need to:

+ Figure out how the "dots" output work for swupd bundle-add and fix commands;
  + Can we know how many dots swupd will print before we start installing?
+ Change the progress calls to use MultiStep

mdhorn:
<a href="https://github.com/mdhorn"><img src="https://avatars.githubusercontent.com/mdhorn" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mdhorn](https://github.com/mdhorn)**
_Thursday May 24, 2018 at 22:00 GMT_

----

Pending updates to swupd
https://github.com/clearlinux/swupd-client/issues/433

mdhorn:
Fixed in release 29750 of Clear Linux* OS

PR #372, opened 2019-05-10
Use the JSON output from swupd to report progress

castulo:
Fixes Issue: #12 

When running the "swupd verify" command to install the base OS and the
selected bundles, the progress of the command execution was just being
shown as a progress bar that was looping until the swupd command
finished. This approach was not ideal since this process can take
several minutes to complete, and since the progress bar was just
spinning it didn't provide a real feedback for the user to know in what
part the process was. This can trigger some anxiety in users.

This commit solves this problem by using the machine readable output
from swupd to determine the exact part of the process where the install is
and report this to the user using a series of steps along with a progress
bar (or percentage value in not TUI installer) that shows how far the
step has gotten.

### DIFF
--- a/swupd/swupd.go
+++ b/swupd/swupd.go
@@ -21,6 +21,8 @@ import (
 	"github.com/clearlinux/clr-installer/log"
 	"github.com/clearlinux/clr-installer/model"
 	"github.com/clearlinux/clr-installer/network"
+	"github.com/clearlinux/clr-installer/progress"
+	"github.com/clearlinux/clr-installer/utils"
 )
 
 var (
@@ -30,6 +32,8 @@ var (
 		"os-core-update",
 		"openssh-server",
 	}
+	prg     progress.Progress
+	prgDesc string
 )
 
 // SoftwareUpdater abstracts the swupd executable, environment and operations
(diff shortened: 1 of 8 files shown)


castulo:
🛑  There is a bug with this code when using the `car-installer-gui`, I'm debugging the problem to provide a fix. Please hold reviews until the bug is fixed.

swupd/swupd.go
```diff
@@ -173,9 +252,10 @@ func (s *SoftwareUpdater) Verify(version string, mirror string, verifyOnly bool)
 
 // VerifyWithBundles runs "swupd verify" operation with all bundles
 func (s *SoftwareUpdater) VerifyWithBundles(version string, mirror string, bundles []string) error {
+	// the "swupd verify" command is being deprecated, use "swupd os-install" instead
 	args := []string{
 		"swupd",
-		"verify",
+		"os-install",
```
otaviobp: I think it's better to leave verify for now on installer and just change it to use os-install when swupd is released with that feature.
If we merge that on installer it won't work with last released version of swupd.

castulo: well, this change cannot be merged in the clr-installer anyway until swupd releases, because it requires some other commits that are in master like the JSON output from swupd. If that is the case the os-install command will be released by then. Unless we are planning on making functionality like the JSON output available sooner than the os-install, like in a different, earlier release.

castulo:
I just rebased this PR and after the rebase I'm seeing some problems with it. I'll take a look at this.

castulo:
Ups, I just figured out what the problem was. I was using the released swupd (3.19.0) in my system which does not have the required changes for this PR to work properly. This PR requires the most recent changes in swupd to work (master). So this PR should not be merged until `swupd 14.0.0` is released. However because of this issue I noticed that when the swupd command fails to run the installer crashes, which should not happen, it should fail gracefully.

castulo:
The PR is now ready for review.

PR #373, opened 2019-05-10
gui: Add vlc as default bundle

reaganlo:
As per Ianeta's request, adding vlc pundle that is selected by default. This is done so that CL videos can be played without crashing.


reaganlo:
Closing this as the requirement has changed.

PR #374, opened 2019-05-13
Add syscheck and localized error dialog

evil-brain:
Run syscheck and report errors after the user chooses their language. This fixes the problem of users booting on Legacy systems and installing on EFI systems to find their system will not boot.

Clear doesn't support Legacy booting beyond the live images, so just return an error and force exiting if EFI isn't found.
Same goes for CPU features, though initrd will refuse to boot if the user is using an ISO image.

Fixes Issue: #232

- Add a syscheck run that executes after clicking 'next' on the localization menu 
- Add a localized dialog box to inform of failures in the chosen language

### DIFF
--- a/gui/window.go
+++ b/gui/window.go
@@ -635,7 +635,7 @@ func (window *Window) launchMenuView() {
 		icon.SetVAlign(gtk.ALIGN_START)
 		contentBox.PackStart(icon, false, true, 0)
 
-		label, err := gtk.LabelNew("System failed to pass pre-install checks.\n\n" + retErr.Error())
+		label, err := gtk.LabelNew(utils.Locale.Get("System failed to pass pre-install checks.") + "\n\n" + retErr.Error())
 		if err != nil {
 			log.Warning("Error creating label")
 			return
@@ -644,7 +644,7 @@ func (window *Window) launchMenuView() {
 		label.SetHAlign(gtk.ALIGN_END)
 		contentBox.PackStart(label, false, true, 0)
 
-		dialog, err := common.CreateDialogOneButton(contentBox, "System Check Failed", utils.Locale.Get("EXIT"), "button-cancel")
+		dialog, err := common.CreateDialogOneButton(contentBox, utils.Locale.Get("System Check Failed"), utils.Locale.Get("EXIT"), "button-cancel")
 		if err != nil {
 			log.Warning("Error creating dialog")
 			return
--- a/gui/window.go
+++ b/gui/window.go
@@ -15,6 +15,7 @@ import (
 	"github.com/clearlinux/clr-installer/log"
 	"github.com/clearlinux/clr-installer/model"
 	"github.com/clearlinux/clr-installer/storage"
+	"github.com/clearlinux/clr-installer/syscheck"
 	"github.com/clearlinux/clr-installer/utils"
 )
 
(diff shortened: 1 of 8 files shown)


gui/window.go
```diff
@@ -603,6 +605,79 @@ func (window *Window) launchWelcomeView() {
...
+	if retErr := syscheck.RunSystemCheck(true); retErr != nil {
+		contentBox, err := gtk.BoxNew(gtk.ORIENTATION_HORIZONTAL, 10)
+		contentBox.SetHAlign(gtk.ALIGN_FILL)
+		contentBox.SetMarginBottom(common.TopBottomMargin)
+		if err != nil {
+			log.Warning("Error creating box")
+			return
+		}
+
+		//Custom CSS class for this dialog - make our error icon red(ish)
```
reaganlo: Do we need the following code. Is it actually setting the icon to the orange color? Can we do this by just creating a corresponding style.css entry?

		//Custom CSS class for this dialog - make our error icon red(ish)
		provider, err := gtk.CssProviderNew()
		if err != nil {
			log.Warning("Error creating CSS provider")
			return
		}
		customTheme := ".dialog image {color: #FDB814;}"
		if err := provider.LoadFromData(customTheme); err != nil {
			log.Warning("Error creating CSS from internal data")
			return
		}

		screen, err := gdk.ScreenGetDefault()
		if err != nil {
			log.Warning("Error getting screen")
			return
		}
		gtk.AddProviderForScreen(screen, provider, gtk.STYLE_PROVIDER_PRIORITY_APPLICATION)

gui/window.go
```diff
@@ -603,6 +605,79 @@ func (window *Window) launchWelcomeView() {
...
+			log.Warning("gtk.ImageNewFromIconName failed for icon dialog-error-symbolic")
+			return
+		}
+
+		icon.SetMarginEnd(12)
+		icon.SetHAlign(gtk.ALIGN_START)
+		icon.SetVAlign(gtk.ALIGN_START)
+		contentBox.PackStart(icon, false, true, 0)
+
+		label, err := gtk.LabelNew("System failed to pass pre-install checks.\n\n" + retErr.Error())
```
reaganlo: Please add entries in the en_US, es_MX and zh_CN locale files for all strings that is displayed in the installer.

gui/window.go
```diff
@@ -603,6 +605,79 @@ func (window *Window) launchWelcomeView() {
...
+		if err != nil {
+			log.Warning("Error creating label")
+			return
+		}
+		label.SetUseMarkup(true)
+		label.SetHAlign(gtk.ALIGN_END)
+		contentBox.PackStart(label, false, true, 0)
+
+		dialog, err := common.CreateDialogOneButton(contentBox, "System Check Failed", utils.Locale.Get("EXIT"), "button-cancel")
```
reaganlo: Please add entries in the en_US, es_MX and zh_CN locale files for all strings that is displayed in the installer.

locale/en_US/LC_MESSAGES/clr-installer.po
```diff
@@ -483,3 +483,24 @@ msgstr "Installation failed."
 
 msgid "Network is not working."
 msgstr "Network is not working."
+
+msgid "System failed to pass pre-install checks."
```
reaganlo: Remove this entry from the en_US file.

locale/en_US/LC_MESSAGES/clr-installer.po
```diff
@@ -483,3 +483,24 @@ msgstr "Installation failed."
...
+msgstr "系统无法通过安装前检查。"
+
+msgid "System Check Failed"
```
reaganlo: Remove this entry from the en_US file.

evil-brain: Hmm, why'd I do that? Will fix these.

gui/window.go
evil-brain: You are correct all this code does is change the icon color for this dialog box (white -> yellow/orange)

I didn't include this in style.css because that would change the icons for all dialog boxes.
Maybe we want that? Or should we define a new type of dialog (informational vs error dialogs)?

reaganlo: Yeah.. To keep it consistent with rest of the application, can we please have this as a style.css entry... Something like ".dialog-warning"

reaganlo: But on a separate note, while testing I still see the icon as Red color and not Orange. Are you able to see it as Orange?

evil-brain: Fixed - the color is the same as the .label-warning in the css:
`#FDB814`

evil-brain:
@mdhorn Fixed the exit path as we discussed

mdhorn:
@reaganlo Are you okay with the updates?

PR #413, opened 2019-05-23
Update Makefile to avoid GRS check

reaganlo:
Update Makefile to avoid GRS check so that it does not behave differently if the source code is in the actual GOPATH.


reaganlo:
Cool.. I'll merge it when you approve it..

Issue #391, opened 2019-05-17
Installer "stalls" after selecting language choice


mdhorn:
**Describe the bug**
When using the graphical Clear Linux* OS Installer from the live-desktop image, the system "stalls" for a few seconds before moving to the next screen.

**To Reproduce**
1. Book the live desktop
2. Click the Installer icon
3. Select a language

**Expected behavior**
The installer should immediately move the next screen, or provide feedback on the reason for the delay.

**Additional context**
This is generally a result of having multiple media attached to the system as all devices need to be probe and partitions and content review for usable media during the installation.
The same "stall" occurs when selecting the "Rescan Media" button on the media selection screen.

reaganlo:
Merged to upstream.

PR #415, opened 2019-05-24
Scan Media in the background

reaganlo:
Fixes Issue: #391

In order to avoid delay in loading the installer, scan the media in the
background as a separate go routine.

### DIFF
--- a/gui/pages/disk_config.go
+++ b/gui/pages/disk_config.go
@@ -269,11 +269,11 @@ func newListStoreMedia() (*gtk.ListStore, error) {
 	return store, err
 }
 
-func runScanLoop(scanInfo ScanInfo) {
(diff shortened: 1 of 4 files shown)


mdhorn:
@reaganlo I still see the long delay from Language selection until the main screen loads on my dev-box.

reaganlo:
@mhorn Made the change to get the results of scanning while loading the Disk Config page and not during its creation.

mdhorn:
@reaganlo I can still launch, select language, select installation media, and the chooser list will appear with no content. If I use the "Rescan" it will get properly updated.  Still a race in there somewhere.

reaganlo:
@mdhorn Ok..I haven't encountered it..yet. Looking into it.

PR #431, opened 2019-06-03
travis: Add workaround for swupd issue

mdhorn:
- Workaround to handle a change in swupd in Docker images under non-clear OSes


mdhorn:
@reaganlo @evil-brain We have a work-around for now. 👍 
Still working with the swupd team to understand why we need this in the Docker environment for ubuntu, but this will at least stop blocking us.

Issue #442, opened 2019-06-14
Update Graphical Telemetry Screen to match Text Telemetry


mdhorn:
**Describe the solution you'd like**
The text-based installer Telemetry screen has two specific buttons for "No, opt-out" and "Yes, opt-in" which makes it clear to the user the selection they are making. We would prefer this same choice presented in the graphical installer versus the checkbox approach.

**Additional context**
Having the user choose the clearly labeled button will help ensure they understand their choice.

reaganlo:
Merged to master.

PR #443, opened 2019-06-15
Enable Telemetry using buttons

reaganlo:

Fixes Issue: #442
- Instead of check box, use buttons to enable/disable telemetry.
- Used simple labels "NO" and "YES" instead of "No, opt-out" and "Yes, opt-in" for precise language translations.

### DIFF
--- a/gui/pages/init.go
+++ b/gui/pages/init.go
@@ -60,7 +60,6 @@ type Page interface {
 type Controller interface {
 	ActivatePage(Page)
 	SetButtonState(flags Button, enabled bool)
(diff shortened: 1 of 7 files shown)


reaganlo:
@mdhorn damn how did I miss that.. :| Good catch Mark..

reaganlo:
@mdhorn I intentionally kept it this way to provide more flexibility for each page to customize the button labels as they want to.

reaganlo:
@mdhorn Made all the changes based on our discussion.

PR #445, opened 2019-06-19
Update Telemetry wording

mdhorn:
- Make the text more clear about the data which might be shared
- Change default to "No choice made" to make it more clear that Telemetry must be visited

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -95,6 +95,11 @@ func init() {
 	eventID = fmt.Sprintf("%x", md5.Sum(randData))
 }
 
+// SetUserDefined set the user defined flag
+func (tl *Telemetry) SetUserDefined(userDefined bool) {
+	tl.userDefined = userDefined
+}
+
 // IsUserDefined returns true if the configuration was interactively
 // defined by the user
 func (tl *Telemetry) IsUserDefined() bool {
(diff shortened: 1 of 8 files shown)


mdhorn:
@reaganlo I need to resolve the problem os knowing if the user selected an answer versus a value read from a configuration file.

telemetry/telemetry.go
```diff
@@ -37,26 +37,18 @@ const (
...
-
-See <a href="http://clearlinux.org/features/telemetry">http://clearlinux.org/features/telemetry</a> for more information.
-`
+	// TelemetryAboutURL is the URL to reference for telemetry details
+	TelemetryAboutURL = `https://clearlinux.org/documentation/clear-linux/concepts/telemetry-about`
 
 	// RequestNotice is a common text string to be displayed when enabling
 	// telemetry by default on local networks
-	RequestNotice = "NOTICE: Enabling Telemetry preferred by default on internal networks"
+	RequestNotice = `NOTICE: Enabling Telemetry preferred by default on internal networks`
```
reaganlo: The missing period at the end is preventing the translation of this line.

mdhorn: > The missing period at the end is preventing the translation of this line.

Good catch! Fixed and resubmitted.

Issue #416, opened 2019-05-24
Check Network in GUI


reaganlo:
**Is your feature request related to a problem? Please describe.**
Network should be set up correctly for the installation to be successful. It is a prerequisite for installation. But currently, the user is made aware of this only in the final Install screen after everything has been selected.

**Describe the solution you'd like**
Test if the network is working properly soon after the language is selected. If the network is not working, the user should not be able to proceed to the next screen and must only be able to exit the Installer.

**Describe alternatives you've considered**
A simpler alternative is to at least inform the user in the form of a label in the first screen that a working network is required for installation.

**Additional context**

Issue #456, opened 2019-06-26
Refactor "gui" package dir structure to match "tui" dir structure


reaganlo:
The "gui" package has some sub packages that are leading to duplicate code. Remove the ambiguous sub packages and make the directory structure of the "gui" package similar to the "tui" package.

This will make the code more clean and concise especially before we add more pages to the GUI to make it feature parity with TUI.

PR #458, opened 2019-06-27
Pre-Check page

reaganlo:

Fixes Issue: #416
- After the language is selected, the pre-check page checks if the prerequisites for the installation are met. Currently the checks include system compatibility and network connectivity.
- Fixes the progress failure status to show a different color.
- Refactored to remove redundant code.
- Displays the system check error.

### DIFF
--- a/gui/pages/pre_check.go
+++ b/gui/pages/pre_check.go
@@ -6,23 +6,24 @@ package pages
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/gotk3/gotk3/glib"
 	"github.com/gotk3/gotk3/gtk"
 
-	ctrl "github.com/clearlinux/clr-installer/controller"
+	"github.com/clearlinux/clr-installer/controller"
+	"github.com/clearlinux/clr-installer/errors"
 	"github.com/clearlinux/clr-installer/gui/common"
 	"github.com/clearlinux/clr-installer/log"
 	"github.com/clearlinux/clr-installer/model"
-	"github.com/clearlinux/clr-installer/network"
 	"github.com/clearlinux/clr-installer/progress"
+	"github.com/clearlinux/clr-installer/syscheck"
 	"github.com/clearlinux/clr-installer/utils"
 )
 
-// InstallPage is a specialised page type with no corresponding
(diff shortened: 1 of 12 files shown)


gui/pages/pre_check.go
```diff
@@ -0,0 +1,294 @@
...
+	msg = msg + " " + utils.Locale.Get("Please wait.")
+	page.info.SetText(msg)
+
+	go func() {
+		progress.Set(page)
+		err := ctrl.PreCheck(page.model)
+		if err != nil {
+			text := strings.Split(err.Error(), "\n")[0]
+			text = text + "\n" + utils.Locale.Get("See %s for details.", page.controller.GetOptions().LogFile)
+			text = text + " " + utils.Locale.Get("Exit.")
```
mdhorn: I would drop this "Exit" as part of the string since the "Exit" button is the only option available to them.

gui/pages/pre_check.go
```diff
@@ -0,0 +1,294 @@
...
+			_, err = glib.IdleAdd(func() {
+				page.controller.SetButtonState(ButtonNext, false)
+			})
+			if err != nil {
+				log.ErrorError(err) // TODO: Handle error in a better way
+				return
+			}
+
+		} else {
+			text := utils.Locale.Get("Prerequisites passed. Please proceed.")
```
mdhorn: Similar to above, I feel like "Prerequisites passed." is sufficient and we should drop the "Please proceed." portion.

gui/pages/pre_check.go
reaganlo: I knew you would say this. I was just trying my luck again :)

gui/pages/pre_check.go
reaganlo: I knew you would say this. I was just trying my luck again :)

reaganlo:
@mdhorn Made all the requested changes.
FYI, I've moved func preCheck() and func systemCheck() to the "init.go" file.
Whenever you give me the green signal, I would like to implement #456 where all the sub-packages within the gui package can be removed and all the functions that are not specific to a struct (class) can be moved to a common file "common.go" similar to the "tui".

Issue #384, opened 2019-05-16
Add kernel command line modification to GUI installer


puneetse:
**Is your feature request related to a problem? Please describe.**
The TUI (installer has a section under Advanced Options to modify kernel cmdline boot parameters. It would be nice for the GUI installer to have this for feature parity. 

**Describe the solution you'd like**
Add kernel command line modification to GUI installer as an advanced option

**Additional context**
This is useful for blacklisting modules known to give issues on first boot.

mdhorn:
@puneetse Thank you. Yes, one of the short list of Advance Options which we have in the text-based installer that need forward porting to the graphical.

mdhorn:
Fixed in version 2.1.0 which should land in a new release of Clear Linux* OS in the next 2-3 days.

Issue #406, opened 2019-05-17
Allow alternate kernel selection in Advanced Options in GUI


mdhorn:
**Is your feature request related to a problem? Please describe.**
As a user installing Clear Linux* OS, I would like to be able to select an alternative kernel from the mainline, lastest native kernel.

**Describe the solution you'd like**
Provide an option to select an alternative kernel under the Advanced Options the same as the text-based installer.

**Describe alternatives you've considered**
Run the Text-based installer from the Desktop in a terminal
`sudo clr-installer-tui`
OR
Post installation, add the desired kernel bundle

evil-brain:
This work is also related to #384 - these options will both be on the same advanced page, one that covers kernel configuration.

mdhorn:
Fixed in version 2.1.0 which should land in a new release of Clear Linux* OS in the next 2-3 days.

PR #459, opened 2019-06-28
gui: Kernel options page

evil-brain:
Adds a kernel options page to the GUI. Plugs into the back end in a similar manner to the same page on the TUI.
Has a feature for default kernels based on what's loaded in the model already, this means virtualbox installations will by default choose the correct kernel.

Removing more than one flag is problematic, it seems that we just pass the whole value to CBM and it doesn't split strings on space and remove all args.

Fixes Issue: #384, #406

- Add an advanced menu option to modify kernel-related settings
- Page contains a list of all available kernels, highlights and selects the default, adds text to the default name to indicate default status.
- Localization support is added, but still needs to be properly reviewed and properly vetted.

### DIFF
--- a/gui/pages/kernel_options.go
+++ b/gui/pages/kernel_options.go
@@ -1,0 +1,296 @@
+// Copyright © 2019 Intel Corporation
+//
+// SPDX-License-Identifier: GPL-3.0-only
+
+package pages
+
+import (
+	"github.com/clearlinux/clr-installer/gui/common"
+
+	"github.com/clearlinux/clr-installer/kernel"
+	"github.com/clearlinux/clr-installer/log"
+	"github.com/clearlinux/clr-installer/model"
+	"github.com/clearlinux/clr-installer/utils"
+	"github.com/gotk3/gotk3/gtk"
+	"strings"
+)
+
+// ConfigKernelPage is a page to change kernel installation configuration
+type ConfigKernelPage struct {
+	controller Controller
+	model      *model.SystemInstall
+	box        *gtk.Box
+	addEntry   *gtk.Entry
+	remEntry   *gtk.Entry
+	addLabel   *gtk.Label
+	remLabel   *gtk.Label
+	scroll     *gtk.ScrolledWindow
+	list       *gtk.ListBox
+	data       []*kernel.Kernel
+	selected   *kernel.Kernel
+}
+
+// NewConfigKernelPage returns a new NewConfigKernelPage
+func NewConfigKernelPage(controller Controller, model *model.SystemInstall) (Page, error) {
+	kernelArgsHelp :=
+		utils.Locale.Get("Note: The boot manager tool will first include the \"Add Extra Arguments\" items.")
+	kernelArgsHelp = kernelArgsHelp + "\n" +
+		utils.Locale.Get("Then, the \"Remove Arguments\" items are removed.")
+	kernelArgsHelp = kernelArgsHelp + "\n" +
+		utils.Locale.Get("The final argument list contains the kernel bundle's configured arguments and the ones configured by the user.")
+
+	data, err := kernel.LoadKernelList()
+	if err != nil {
+		return nil, err
+	}
+	page := &ConfigKernelPage{
+		controller: controller,
+		model:      model,
+		data:       data,
+	}
+
+	// Box
+	page.box, err = setBox(gtk.ORIENTATION_VERTICAL, 0, "box-page")
+	if err != nil {
+		return nil, err
+	}
+
+	// ScrolledWindow
+	page.scroll, err = setScrolledWindow(gtk.POLICY_NEVER, gtk.POLICY_AUTOMATIC, "scroller")
+	if err != nil {
+		return nil, err
+	}
+	page.scroll.SetMarginStart(common.StartEndMargin)
+	page.scroll.SetMarginEnd(common.StartEndMargin)
+	page.box.PackStart(page.scroll, true, true, 0)
+
+	// ListBox
+	page.list, err = setListBox(gtk.SELECTION_SINGLE, true, "list-scroller")
+	if err != nil {
+		return nil, err
+	}
+	if _, err := page.list.Connect("row-activated", page.onRowActivated); err != nil {
+		return nil, err
+	}
+	page.scroll.Add(page.list)
+
+	// Create list data
+	for _, v := range page.data {
+		name := v.Name
+		desc := utils.Locale.Get(v.Desc)
+
+		//Add default label to kernel already in model
+		if page.model.Kernel != nil {
+			if v.Bundle == page.model.Kernel.Bundle {
+				name = v.Name + utils.Locale.Get(" (default)")
+			}
+		}
+
+		box, err := setBox(gtk.ORIENTATION_VERTICAL, 0, "box-list-label")
+		if err != nil {
+			return nil, err
+		}
+
+		labelName, err := setLabel(name, "list-label-name", 0.0)
+		if err != nil {
+			return nil, err
+		}
+		box.PackStart(labelName, false, false, 0)
+
+		labelDesc, err := setLabel(desc, "list-label-desc", 0.0)
+		if err != nil {
+			return nil, err
+		}
+		box.PackStart(labelDesc, true, true, 0)
+
+		page.list.Add(box)
+	}
+
+	// addLabel: Tell users they can add kernel command lines below
+	addText := utils.Locale.Get("Add Extra Arguments")
+
+	page.addLabel, err = setLabel(addText, "label-entry", 0.0)
+	if err != nil {
+		return nil, err
+	}
+	page.addLabel.SetMarginStart(common.StartEndMargin)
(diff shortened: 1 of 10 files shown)


gui/pages/kernel_options.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// GetID returns the ID for this page
+func (page *ConfigKernelPage) GetID() int {
+	return PageIDConfigKernel
+}
+
+// GetIcon returns the icon for this page
+func (page *ConfigKernelPage) GetIcon() string {
+	return "computer"
```
mdhorn:
```suggestion
	return "applications-engineering"
```

mdhorn: Please update to a unique icon; I like this one best for now.

locale/en_US/LC_MESSAGES/clr-installer.po
```diff
@@ -543,6 +543,33 @@ msgstr "Installation in progress."
...
+msgid "Add Extra Arguments:"
+msgstr "Add Extra Arguments:"
+
+msgid "Remove Arguments:"
+msgstr "Remove Arguments:"
+
+msgid "Kernel Configuration"
+msgstr "Kernel Configuration"
+
+msgid ", custom command line arguments"
```
mdhorn: Please drop the comma from the string. Instead, let's have two sentences.

locale/en_US/LC_MESSAGES/clr-installer.po
```diff
@@ -543,6 +543,33 @@ msgstr "Installation in progress."
 msgid "Installation successful."
 msgstr "Installation successful."
 
+msgid " (default)"
+msgstr " (default)"
+
+msgid "Add Extra Arguments:"
+msgstr "Add Extra Arguments:"
```
mdhorn: Please drop the : from the label strings to be consistent with the User Add screen

gui/pages/kernel_options.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+		box.PackStart(labelDesc, true, true, 0)
+
+		page.list.Add(box)
+	}
+
+	// addLabel: Tell users they can add kernel command lines below
+	addText := utils.Locale.Get("Add Extra Arguments:")
+
+	page.addLabel, err = setLabel(addText, "label-addLabel", 0.0)
```
mdhorn: Please change the style to the "label-entry" which we are using for the add user screen for consistency.

gui/pages/kernel_options.go
```diff
@@ -0,0 +1,292 @@
...
+		return nil, err
+	}
+	page.addEntry.SetMarginStart(common.StartEndMargin)
+	page.addEntry.SetMarginEnd(common.StartEndMargin)
+	page.addEntry.SetTooltipText(utils.Locale.Get(kernelArgsHelp))
+	page.box.PackStart(page.addEntry, false, false, 0)
+
+	// remLabel label
+	remText := utils.Locale.Get("Remove Arguments:")
+	page.remLabel, err = setLabel(remText, "label-remLabel", 0.0)
```
mdhorn: Please change the style to the "label-entry" which we are using for the add user screen for consistency.

gui/pages/kernel_options.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+	if _, err := page.list.Connect("row-activated", page.onRowActivated); err != nil {
+		return nil, err
+	}
+	page.scroll.Add(page.list)
+
+	// Create list data
+	for _, v := range page.data {
+		name := v.Name
+		desc := v.Desc
```
mdhorn: Please add a call to the util localization to the description of the kernel and add entries to the po files.
The name may not make sense to translate, but if we leave that in English having the description translated would be helpful.

evil-brain: Done.

gui/pages/kernel_options.go
evil-brain: Done.

gui/pages/kernel_options.go
evil-brain: Done.

locale/en_US/LC_MESSAGES/clr-installer.po
evil-brain: Done.

locale/en_US/LC_MESSAGES/clr-installer.po
evil-brain: Done

gui/pages/kernel_options.go
evil-brain: Done.

evil-brain:
> I can't seem to clear the kernel arguments. I would expect if I clear the field and Confirm that it would be removed. Also, if I clear and put something new, it is added to the previous value.

Fixed this in the next push.

Also, I guess I should add a massive banner to my motd or maybe a commit hook:
```
    ____  __  ___   __   ________  ________   __    _____   __________________  ______
   / __ \/ / / / | / /  /_  __/ / / / ____/  / /   /  _/ | / /_  __/ ____/ __ \/ / / /
  / /_/ / / / /  |/ /    / / / /_/ / __/    / /    / //  |/ / / / / __/ / /_/ / / / / 
 / _, _/ /_/ / /|  /    / / / __  / /___   / /____/ // /|  / / / / /___/ _, _/_/_/_/  
/_/ |_|\____/_/ |_/    /_/ /_/ /_/_____/  /_____/___/_/ |_/ /_/ /_____/_/ |_(_|_|_)   

```

mdhorn:
> There is a small bug right now where if you click the "Confirm" button on the kernel screen with no values for the command line it crashes. I've emailed you a patch; which will also fix the text-based user interface.

In case the email didn't work, here is a gist
https://gist.github.com/mdhorn/10155812faf1464df9857c124359f349

evil-brain:
Applied/tested Mark's patch, it fixes the issue for me! Thanks, Mark!

gui/pages/kernel_options.go
```diff
@@ -0,0 +1,296 @@
...
+				name = v.Name + utils.Locale.Get(" (default)")
+			}
+		}
+
+		box, err := setBox(gtk.ORIENTATION_VERTICAL, 0, "box-list-label")
+		if err != nil {
+			return nil, err
+		}
+
+		labelName, err := setLabel(name, "list-label-name", 0.0)
```
reaganlo: I don't think this style "list-label-name" is defined.

gui/pages/kernel_options.go
```diff
@@ -0,0 +1,296 @@
...
+		if err != nil {
+			return nil, err
+		}
+		box.PackStart(labelName, false, false, 0)
+
+		labelDesc, err := setLabel(desc, "list-label-desc", 0.0)
```
reaganlo: I don't think this style "list-label-desc" is defined.

gui/pages/kernel_options.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	// Box
+	page.box, err = setBox(gtk.ORIENTATION_VERTICAL, 0, "box-page")
+	if err != nil {
+		return nil, err
+	}
+
+	// ScrolledWindow
+	page.scroll, err = setScrolledWindow(gtk.POLICY_NEVER, gtk.POLICY_AUTOMATIC, "scroller")
```
reaganlo: Can we have a heading for this? Like "Select Kernel"

gui/pages/kernel_options.go
```diff
@@ -0,0 +1,296 @@
...
+	page.addLabel, err = setLabel(addText, "label-entry", 0.0)
+	if err != nil {
+		return nil, err
+	}
+	page.addLabel.SetMarginStart(common.StartEndMargin)
+	page.addLabel.SetHAlign(gtk.ALIGN_START)
+	page.box.PackStart(page.addLabel, false, false, 10)
+
+	// addEntry: Args to add to the kernel command line
+	page.addEntry, err = setEntry("entry")
```
reaganlo: This style "entry" is adding an extra top padding causing additional space between heading and the entry. Can we remove this additional space by defining another style for this?

gui/pages/kernel_options.go
```diff
@@ -0,0 +1,296 @@
...
+	page.remLabel, err = setLabel(remText, "label-entry", 0.0)
+	if err != nil {
+		return nil, err
+	}
+	page.remLabel.SetMarginStart(common.StartEndMargin)
+	page.remLabel.SetMarginEnd(common.StartEndMargin)
+	page.box.PackStart(page.remLabel, false, false, 10)
+
+	// remEntry: Args to remove to the kernel command line.
+	page.remEntry, err = setEntry("entry")
```
reaganlo: This style "entry" is adding an extra top padding causing additional space between heading and the entry. Can we remove this additional space by defining another style for this?

evil-brain:
Thanks for the patch @reaganlo!

At this point, all 3 of us worked on this, I think we can almost call this "crowd sourced source"  :stuck_out_tongue_winking_eye:

Issue #417, opened 2019-05-24
Catch Fails and display to user in GUI


mdhorn:
**Describe the solution you'd like**
Catch errors and display the reason allow with helpful information for opening an issue similar to what the TUI does on fails.

reaganlo:
@mdhorn What should be the title and msg body of the popup?

Should the msg body be this:
	fmt.Println("Please report this crash using GitHub Issues:")
	fmt.Println("\thttps://github.com/clearlinux/clr-installer/issues")

	fmt.Println("")

	fmt.Println("Include the following as attachments to enable diagnosis:")
	fmt.Printf("\t%s\n", preConfName)
	fmt.Printf("\t%s\n", logFileName)

	fmt.Println("")

	fmt.Println("If the problem persists, enabling additional logging will be helpful in")
	fmt.Println("diagnosing the issue. At the OS Boot screen, hit 'e' to edit the kernel")
	fmt.Println("command line. Add the following to the end of the line:")
	fmt.Println("\tclri.loglevel=4")

mdhorn:
Fixed in version 2.1.0 which should land in a new release of Clear Linux* OS in the next 2-3 days.

PR #460, opened 2019-07-02
Handle crashes

reaganlo:
Fixes Issue: #417

- Introduces Panic() to handle crashes
- Removed "model" and "installReboot" variables from Gui struct as they are not being used.
- Updated log level from warning to errors wherever applicable.

### DIFF
--- a/gui/window.go
+++ b/gui/window.go
@@ -820,7 +820,7 @@ func displayErrorDialog(err error) {
 		return
 	}
 
-	label, err := gtk.LabelNew(log.GetCrashInfoMsg())
+	label, err := common.SetLabel(log.GetCrashInfoMsg(), "label-error", 0.0)
 	if err != nil {
 		log.Error("Error creating label", err)
 		return
--- a/gui/window.go
+++ b/gui/window.go
@@ -655,7 +655,7 @@ func (window *Window) launchWelcomeView() {
 	window.mainLayout.Remove(window.contentLayout)
 	window.mainLayout.Remove(window.banner.GetRootWidget())
 	if _, err := window.createWelcomePage(); err != nil {
-		log.ErrorError(err) // TODO: Handle error
+		window.Panic(err)
 	}
 }
 
@@ -665,7 +665,7 @@ func (window *Window) launchPreCheckView() {
 	window.menu.currentPage.StoreChanges()
 
 	if _, err := window.createPreCheckPage(); err != nil {
-		log.ErrorError(err) // TODO: Handle error
+		window.Panic(err)
 	}
 }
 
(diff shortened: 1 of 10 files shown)


gui/window.go
```diff
@@ -781,6 +782,16 @@ func (window *Window) SetScanMedia(scannedMedia []*storage.BlockDevice) {
 	window.scanInfo.Media = scannedMedia
 }
 
+// Panic handles the gui crashes
+func (window *Window) Panic(err error) {
+	log.Debug("Panic")
+	displayErrorDialog(err)
```
mdhorn: Should this move after the log.RequestCrashInfo()? To ensure we log the telemetry and the crash info before the user has a chance to dismiss and close the app?

gui/window.go
```diff
@@ -800,3 +811,39 @@ func GetThankYouMessage() string {
...
+	if err != nil {
+		log.Error("Error creating label", err)
+		return
+	}
+	label.SetHAlign(gtk.ALIGN_CENTER)
+	label.SetMarginBottom(common.TopBottomMargin)
+	label.SetSelectable(true)
+	contentBox.PackStart(label, true, true, 0)
+
+	title := utils.Locale.Get("Sorry, something went wrong...")
```
mdhorn: Maybe we just say "Something went wrong"? Debating the "Sorry" portion.

log/log.go
```diff
@@ -86,26 +87,23 @@ func SetOutputFilename(logFile string) (*os.File, error) {
 	return filehandle, nil
 }
 
+// GetCrashInfoMsg returns the crash info message.
+func GetCrashInfoMsg() string {
+	msg := utils.Locale.Get("Please report this crash using %s", "GitHub Issues:")
+	msg += "\n" + "https://github.com/clearlinux/clr-installer/issues"
+	msg += "\n\n" + "Include the following as attachments to enable diagnosis:"
```
mdhorn: Should we not localize all of these sentences?

log/log.go
```diff
@@ -86,26 +87,23 @@ func SetOutputFilename(logFile string) (*os.File, error) {
...
+	msg += "\n" + preConfName
+	msg += "\n" + logFileName
+	msg += "\n\n" + "You may need to remove any personal data of concern from the attachments."
```
mdhorn: Should we not localize all of these sentences?

gui/window.go
reaganlo: Done.

gui/window.go
reaganlo: Done.

log/log.go
reaganlo: Yeah we should. Completely missed out the translations during rebase.
Good catch.

log/log.go
reaganlo: Done.

reaganlo:
@mdhorn I just modified the code to manually return an errors.Errorf("TEST") when err == nil where Panic is called.
Not a perfect test.

reaganlo:
@mdhorn Made the changes for error popup font color and pre-check page translation fix.

Issue #405, opened 2019-05-17
Add "Disable Autoupdate" to Advanced Options in GUI


mdhorn:
**Describe the solution you'd like**
Provide an option under Advanced to Disable Automatic Updates the same as the Text-based UI.

**Additional context**
This is useful for debugging or recreating an issue under a specific release.

mdhorn:
Fixed in version 2.1.0 which should land in a new release of Clear Linux* OS in the next 2-3 days.

Issue #455, opened 2019-06-26
UserAdd page increases size of Installer window


reaganlo:
**Describe the bug**
The UserAdd page increases the width of the Installer window.

**To Reproduce**
Launch Installer and visually note the size of the window. Now go to the Menu page by clicking Next. You will notice the width of the window increases.

Now comment out the [code](https://github.com/clearlinux/clr-installer/blob/master/gui/window.go#L259) that creates the UserAdd page and try the same. You will notice that the width of the window remains constant.

**Expected behavior**
At least when translation is not applied (i.e for the English language), the Installer window size should remain constant. Currently the UserAdd page has lot of unnecessary real estate that can be made more concise.

Issue #407, opened 2019-05-17
Add swupd mirror to the Advanced Options in GUI


mdhorn:
**Is your feature request related to a problem? Please describe.**
As a Clear Linux* OS installer I would like the option of specifying an alternative mirror from where to download via swupd.

**Describe the solution you'd like**
Add an advanced option to enter the Swupd Mirror URL the same as the Text-based installer.

**Describe alternatives you've considered**
Run the TUI from the Desktop in a terminal:
`sudo clr-installer-tui`
OR
Launch the GUI from a terminal and specify the mirror URL:
`sudo clr-installer-gui --swupd-mirror <URL>`

mdhorn:
Consider implementing on the same Advance Options screen as #228

reaganlo:
Done and merged to upstream.

mdhorn:
Fixed in version 2.1.0 which should land in a new release of Clear Linux* OS in the next 2-3 days.

PR #464, opened 2019-07-09
gui: Introduces swupd config page

reaganlo:

Fixes Issue: #405 #407 #455

- Introduces swupd config page that can be used to set Mirror URL and Auto Updates
- Moved associated common GUI and TUI strings to swupd.go
- Added corresponding translations
- Made the change to display swupd related pages in sequence in the TUI menu.
- Fixed resize issue due to long strings in **User Add** page #455 as part of fixing the same for this page.
- Refactored code to handle closing pages due to a bug introduced in this PR.

### DIFF
--- a/gui/pages/swupd_config.go
+++ b/gui/pages/swupd_config.go
@@ -67,6 +67,8 @@ func NewSwupdConfigPage(controller Controller, model *model.SystemInstall) (Page
 		return nil, err
 	}
 	page.mirrorDesc.SetMarginStart(common.StartEndMargin)
+	page.mirrorDesc.SetMaxWidthChars(1) // The value does not matter but its required for LineWrap to work
+	page.mirrorDesc.SetLineWrap(true)
 	page.box.PackStart(page.mirrorDesc, false, false, 10)
 
 	page.mirrorEntry, err = setEntry("entry-no-top-margin")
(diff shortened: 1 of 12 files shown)


swupd/swupd.go
```diff
@@ -412,7 +459,7 @@ func setMirror(swupdArgs []string, t string) (string, error) {
 func SetHostMirror(url string) (string, error) {
 
 	if urlErr := network.CheckURL(url); urlErr != nil {
-		return "", fmt.Errorf("Server not responding")
+		return "", fmt.Errorf(utils.Locale.Get("Server not responding"))
```
mdhorn: Initially, I had concerns about having these translated in common code as this means the TUI will get translated strings that might now format right for other languages. But looking through the code base we have similar issues elsewhere, so we will not worry about this for now.

gui/pages/swupd_config.go
```diff
@@ -0,0 +1,284 @@
...
+func (page *SwupdConfigPage) GetConfiguredValue() string {
+	var ret string
+	if page.model.AutoUpdate {
+		ret = utils.Locale.Get("Auto updates enabled")
+	} else {
+		ret = utils.Locale.Get("Auto updates disabled")
+	}
+
+	if page.model.SwupdMirror != "" {
+		ret += utils.Locale.Get(". Custom mirror set.")
```
mdhorn: I'm not sure that it makes sense to have this leading person here for the previous statement.

reaganlo:
@mdhorn As discussed, I've made the following changes:
- Added corresponding translations
- Made the change to display swupd related pages in sequence in the TUI menu.
- Fixed resize issue due to long strings in **User Add** page #455 as part of fixing the same for this page.
- Refactored code to handle closing pages due to a bug introduced in this PR.

gui/pages/swupd_config.go
reaganlo: Done to have "." as a separate translated string.

PR #466, opened 2019-07-10
Reduce bundles in hyperv.yml

puneetse:
Fixes Issue: https://github.com/clearlinux/distribution/issues/622

- Remove  editors, network-basic, and sysadmin-basic bundle from Hyper-V image.

> This should be user-installed, not pre-installed in my opinion. Especially without a good way to recursively remove bundles.

### DIFF
--- a/scripts/hyperv.yaml
+++ b/scripts/hyperv.yaml
@@ -28,12 +28,9 @@ targetMedia:
 
 bundles: [
     bootloader,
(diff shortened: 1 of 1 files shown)


mdhorn:
@puneetse The yaml files in the scripts directory are for reference.
DevOps has there own repository of YAML files they use to generate the images they publish.

PR #472, opened 2019-07-16
Execute deferred functions during failure

### LINKED ISSUE #463: Clean lock file when external processes exit with non-zero status
**Describe the bug**
When clr-installer is calling out to external processes, be it post install scripts, swupd, telem-record-gen, etc, if one of these processes exits with a non-zero status, the .lock file that the clr-installer creates lingers around and is not removed.  This should be removed since the clr-installer has a chance to clean this up.

**To Reproduce**
1. Create an image with a disk size that is too small to trigger swupd to run out of disk space
2. See that the clr-installer exits and leaves around it's corresponding .lock file.

**Expected behavior**
The clr-installer should remove the .lock file when once of its external processes exits with a non-zero status.

**Screenshots**
N/A

**Environment (please complete the following information):**
 - Clear Linux OS Version: N/A
 - Installer Version: 2.0.8
 - Platform: N/A

**Additional context**
N/A

jwakre:
Fixes Issue: #463 

- Replace os.Exit with fatal
- Add error channel for goroutines in main.go
- Remove unused unlock code from fatal function

### DIFF
--- a/clr-installer/main.go
+++ b/clr-installer/main.go
@@ -43,13 +43,6 @@ var (
 )
 
 func fatal(err error) {
-	if lock != "" {
-		lErr := lock.Unlock()
-		if lErr != nil {
-			fmt.Printf("Cannot lock %q, reason: %v\n", lock, lErr)
-		}
-	}
-
 	log.ErrorError(err)
 	panic(err)
 }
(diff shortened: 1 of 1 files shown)


clr-installer/main.go
```diff
@@ -43,13 +43,6 @@ var (
 )
 
 func fatal(err error) {
-	if lock != "" {
```
jwakre: One of the deferred functions will unlock the lock. Also, the global lock value is never set, so this condition will never be true (The lock value is set with a local scope, so it does not set the global scope).

Issue #473, opened 2019-07-16
Verify swupd URLs are HTTPS (or file)


mdhorn:
**Describe the bug**
The new release of swupd-client (3.21.0) now requires that URLs be HTTPS (or file) as HTTP is considered insecure. The change causes the installer to fail if HTTP URL is used for the content or version URLs.

**To Reproduce**
1. Build an image file using one of the example YAML configuration files and also supply the `--swupd-contenturl <HTTP_URL>`

**Expected behavior**
The installer should provide a useful message that HTTP is not supported.

**Additional context**
The same validation check of the URL to ensure it is either HTTPS or FILE needs to be done for the user input field for the swupd mirror in the TUI and GUI.

mdhorn:
Fixed in release 2.1.1  which should land in Clear Linux* OS release > 30610

PR #474, opened 2019-07-17
swupd: Require HTTPS/FILE protocol for mirror

jwakre:
Fixes Issue: #473

- Return error when swupd-contenturl argument does not use HTTPS or FILE protocol
- Prevent gui/tui from accepting URL that is not the HTTPS or FILE protocol
- Change invalid URL string for gui/tui to suggest HTTPS: "Invalid URL" -> "Invalid URL: Use HTTPS"
- Update locales **(I am not confident in my Chinese translation, so I did not update this)**
- Mixed/Uppercase HTTP and FILE protocols are accepted (Swupd accepts Mixed/Uppercase)

### DIFF
--- a/locale/zh_CN/LC_MESSAGES/clr-installer.po
+++ b/locale/zh_CN/LC_MESSAGES/clr-installer.po
@@ -655,8 +655,8 @@ msgstr "启用自动更新"
 msgid "WARNING: Disabling Automatic OS Updates puts the system at risk of missing critical security patches."
 msgstr "警告: 禁用自动操作系统更新会使系统面临丢失关键安全修补程序的风险。"
 
-msgid "Invalid URL"
-msgstr "无效 URL"
+msgid "Invalid URL: Use HTTPS"
+msgstr "无效 URL: 使用 HTTPS"
 
 msgid "Mirror not set correctly"
 msgstr "镜像设置不正确"
--- a/swupd/swupd.go
+++ b/swupd/swupd.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -66,7 +67,7 @@ const (
 	AutoUpdateWarning2 = "missing critical security patches."
 
 	// InvalidURL specifies invalid url error message
-	InvalidURL = "Invalid URL"
+	InvalidURL = "Invalid URL: Use HTTPS"
 
 	// IncorrectMirror specifies incorrect mirror error message
 	IncorrectMirror = "Mirror not set correctly"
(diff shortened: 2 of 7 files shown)


locale/zh_CN/LC_MESSAGES/clr-installer.po
```diff
@@ -655,7 +655,7 @@ msgstr "启用自动更新"
 msgid "WARNING: Disabling Automatic OS Updates puts the system at risk of missing critical security patches."
 msgstr "警告: 禁用自动操作系统更新会使系统面临丢失关键安全修补程序的风险。"
 
-msgid "Invalid URL"
+msgid "Invalid URL: Use HTTPS"
```
jwakre: My online translation was much longer than the existing translation, so I did not change it.

reaganlo: Sorry I did not understand this. Why are we not having the full Chinese translation **无效 URL:使用 HTTPS**

swupd/swupd.go
```diff
@@ -66,7 +67,7 @@ const (
 	AutoUpdateWarning2 = "missing critical security patches."
 
 	// InvalidURL specifies invalid url error message
-	InvalidURL = "Invalid URL"
+	InvalidURL = "Invalid URL: Use HTTPS"
```
reaganlo: @mdhorn Is it ok that we are not explicitly mentioning **FILE** as one of the allowed URI schemes for the swupd mirror?

swupd/swupd.go
```diff
@@ -525,6 +526,22 @@ func UnSetHostMirror() (string, error) {
...
+// IsValidMirror checks for valid URIs that use the HTTPS or FILE protocol
+func IsValidMirror(mirror string) bool {
+	_, err := url.ParseRequestURI(mirror)
+	if err != nil {
+		return false
+	}
+
+	httpPrefix := strings.HasPrefix(strings.ToLower(mirror), "https:")
+	filePrefix := strings.HasPrefix(strings.ToLower(mirror), "file:")
+	if httpPrefix != true && filePrefix != true {
```
reaganlo: @mdhorn @otaviobp So we are going to limit the user to using only **HTTPS** and **FILE** URI schemes as the swupd mirror in the Installer. Is that OK?

locale/zh_CN/LC_MESSAGES/clr-installer.po
jwakre: I didn't trust my translation: 網址無效：使用HTTPS. I can copy the one from your post.

swupd/swupd.go
```diff
@@ -525,6 +526,22 @@ func UnSetHostMirror() (string, error) {
...
 }
 
+// IsValidMirror checks for valid URIs that use the HTTPS or FILE protocol
+func IsValidMirror(mirror string) bool {
+	_, err := url.ParseRequestURI(mirror)
+	if err != nil {
+		return false
+	}
+
+	httpPrefix := strings.HasPrefix(strings.ToLower(mirror), "https:")
```
jwakre: s/http/https

jwakre:
This PR has been updated with the following changes:
* Added Chinese translation
* Renamed httpPrefix to httpsPrefix

It was decided that the installer will suggest HTTPS, but it will also accept FILE. Other protocols will not be accepted.

jwakre:
There are concerns about swupd's restrictions for the HTTP protocol. I am going to hold-off on merging this PR until these concerns are resolved.

jwakre:
> There are concerns about swupd's restrictions for the HTTP protocol. I am going to hold-off on merging this PR until these concerns are resolved.

This PR is acceptable, but support for HTTP mirrors is still desired. I created #478 to add HTTP mirror support. Users will be required to acknowledge (a checkbox or flag) that they are enabling insecure downloads to enable HTTP.

Issue #237, opened 2019-02-26
Partition Ordering


mdhorn:
**Describe the bug**
The partition table is created incorrectly if the partitions are not in sorted order.

**To Reproduce**
1. Take the released scripts/installer.yaml file and change the order of the partition; move p3 to the first child position
2. Attempt to build image
3. Image build fails due to lack of disk space.  The root partition was placed in the wrong location and is the size of the swap.

**Expected behavior**
The image is built correctly with the request partition table.

**Environment (please complete the following information):**
 - Clear Linux OS Version: Installed version: 27800
 - Installer Version 1.1.5

mdhorn:
The code is making the assumption that the partitions are in order of the partition table.
Need to either sort them to ensure this or modify the code to remove the dependency.
`func (bd *BlockDevice) WritePartitionTable(legacyBios bool) error {`

mdhorn:
@reaganlo We are seeing this bug show up when using the JSON -> YAML converter functionality; 50% of the time converting the same JSON file results in a YAML file with the partitions not in ascending order.

reaganlo:
@mdhorn Yes Mark. That is correct. This was discovered during the json to yaml feature implementation. At that point we decided that the correct fix would be to make the installer smart to read the yaml tags and understand the type of partition during installation and not go by the sequence in which they appear in the config file.

Is there a different action item now?

mdhorn:
@reaganlo This is just bubbling up in importance due to the increased usage of the clr-installer using JSON from ister.

reaganlo:
Fixed and merged to master.

mdhorn:
Fixed in release 2.1.1  which should land in Clear Linux* OS release > 30610

PR #481, opened 2019-07-23
storage: Order partitions by name

reaganlo:
Sort partitions by name in ascending order before writing the partition tables. 
Fixes issue: #237

Signed-off-by: Reagan Lopez <reagan.lopez@intel.com>


mbelluzzo:
I think it is still something wrong:

https://gist.github.com/mbelluzzo/a61c4cd16f2fe4d0dd9031a41e0efc24

reaganlo:
@mbelluzzo The fix is not to order the partitions in the config files. But to ensure that while writing the partition tables during install, they are done in the order of partition name and installation does not fail.

mbelluzzo:
Oh, fine hehehe I thought that it was on the conversion tool. My bad ;-)

PR #486, opened 2019-07-31
Remove bug in readme example

mdhorn:
The shell will not expand a `~` when it is proceeded
with an equal sign (`=`). To use `~` in a path, then be
sure to have it preceded by a space.

Fixes: 485

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -101,6 +101,6 @@ sudo .gopath/bin/clr-installer --reboot=false
 or if using the Mass Installer mode:
 
 ```
-sudo .gopath/bin/clr-installer --config=~/my-install.yaml --reboot=false
+sudo .gopath/bin/clr-installer --config ~/my-install.yaml --reboot=false
 ```
 


lebensterben:
@mdhorn Shell will first perform filename expansion before passing the arguments to the command `sudo`.

See this example

```
❯ sudo go run main.go -strFlag ~ 
strflag has value  /home/lucius

```

where the main.go has contents as

```go
 package main
 
 import "fmt"
 import "flag"
 
 var strflag *string = flag.String("strFlag","s","help for strflag")
 
 func main() {
   flag.Parse()
     fmt.Println("strflag has value ", *strflag)
 }

```

mdhorn:
@lebensterben It is definitely the SHELL not expanding ~ when it is preceeded by =

bash:
```
$ sudo echo clr-installer --config=~/cloudguest.yaml
clr-installer --config=~/cloudguest.yaml
$ sudo echo clr-installer --config ~/cloudguest.yaml
clr-installer --config /home/mhorn/cloudguest.yaml
$ sudo echo clr-installer -c=~/cloudguest.yaml
clr-installer -c=~/cloudguest.yaml
$ sudo echo clr-installer -c ~/cloudguest.yaml
clr-installer -c /home/mhorn/cloudguest.yaml
```

zsh:
```
% sudo echo clr-installer --config=~/cloudguest.yaml
clr-installer --config=~/cloudguest.yaml
% sudo echo clr-installer --config ~/cloudguest.yaml
clr-installer --config /home/mhorn/cloudguest.yaml
% sudo echo clr-installer -c=~/cloudguest.yaml
clr-installer -c=~/cloudguest.yaml
% sudo echo clr-installer -c ~/cloudguest.yaml
clr-installer -c /home/mhorn/cloudguest.yaml
```

lebensterben:
>@lebensterben It is definitely the SHELL not expanding ~ when it is preceeded by = ...

@mdhorn Thanks for the clarification. Indeed having ~ directory are problematic.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

lebensterben:
@mdhorn 
I'm writing shell completions for `clr-installer` and let me know whether the following is okay.

Although having ~ as option argument is discouraged, it doesn't behave well only when it follows an equal sign. In `bash` the `=` is not inserted automatically, and in `zsh` I can make it not appending a `=` to a long option. Then filename expansion would behave as expected.

Alternatively, in `zsh` I can print a message to ask the user to avoid `~` when it follows an equal sign, and in `bash` I can replace it with it's real path by the completion system.

What do you think?

mdhorn:
> @mdhorn
> I'm writing shell completions for `clr-installer` and let me know whether the following is okay.
> 
> Although having ~ as option argument is discouraged, it doesn't behave well only when it follows an equal sign. In `bash` the `=` is not inserted automatically, and in `zsh` I can make it not appending a `=` to a long option. Then filename expansion would behave as expected.
> 
> Alternatively, in `zsh` I can print a message to ask the user to avoid `~` when it follows an equal sign, and in `bash` I can replace it with it's real path by the completion system.
> 
> What do you think?

For the file completion, I think it should always generate working syntax for the paths. So your first option sounds best.

Issue #484, opened 2019-07-27
Failure to launch live-desktop


nsgnn:
Machine is already running a clear linux install, but I cannot launch a live-desktop version of clear linux due to `failing to find installer media...` My main disk is encrpyted with luks, but beyond that it is the same hardware as when I previously installed clear linux using a live desktop image. The image is fresh off the website though. I followed the "create a boot usb" tutorial on the clear linux site.

![Error screenshot](https://user-images.githubusercontent.com/978752/62055931-374c8e00-b1e2-11e9-8948-c099c9394c33.png)

mdhorn:
The problem appears to be in the recent releases of Clear Linux. 
The initrd process is not able to find the installer media likely due to a recent kernel change in the distribution. 
The problem is under investigation.

miguelinux:
@mdhorn is the same for the CLEAR-4703

mdhorn:
Fixed in release 2.1.1  which should land in Clear Linux* OS release > 30610

nsgnn:
This issue persisted with version 30650 on host machine and usb version(the one failing to boot to live-desktop).

mdhorn:
@nsgnn The fix is was how the ISO image was created; not in the host OS.
I just download and boot the 30650 ISO image in qemu
`qemu-system-x86_64 -enable-kvm -cpu host -smp 4 -m 1024 -cdrom clear-30650-live-server.iso`

If this image is not booting for you, please provide the full command you are using or verify the checksum on the download ISO.

EdPhere:
I'm getting the Failed to find installer media when booting the iso which was downloaded from here.  I've tried every bootparm I know of to get it to recognize the iso.

Grub2Win menu

menuentry " Clear Linux* OS"   --class other --class icon-gnugrub  {

     set lang="en" 
     set iso="/Users/Ed/Downloads/clear-32270-live-desktop.iso"
     set bootparms="from=all, from=$iso, fromiso=$iso, findiso=$iso, iso-scan/filename=$iso, bootfrom=/dev/sda6$iso    qemu-system-x86_64 -enable-kvm -cpu host -smp 4 -m 1024 -cdrom $iso"

     set bootdrv=$root
     search -f $iso --set=root 
     if [ $? = 0 ] ; then
        loopback loop $iso
        linux  (loop)/kernel/kernel.xz $bootparms
        initrd (loop)/EFI/BOOT/initrd.gz 
        loopback --delete loop
     else
[...]

PR #487, opened 2019-07-31
isoutils: fix mkInitrd install os-core-plus

miguelinux:
Since clearlinux/clr-bundles@28319f78d1bd ("os-core: Remove content
provided by os-core-plus"), **kmod** package was removed from
os-core bundle and it is shipped by os-core-plus.

Kmod package provides insmod which is used in the ISO initrd/init
script.

Use **os-core-plus** bundle as initrd base instead of os-core.

### DIFF
--- a/isoutils/isoutils.go
+++ b/isoutils/isoutils.go
@@ -123,8 +123,8 @@ func mkInitrd(version string, model *model.SystemInstall, options args.Args) err
 	options.SwupdFormat = "staging"
 	sw := swupd.New(tmpPaths[clrInitrd], options)
 
-	/* Install os-core only as initrd */
-	if err := sw.VerifyWithBundles(version, model.SwupdMirror, []string{}); err != nil {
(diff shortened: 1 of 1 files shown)


miguelinux:
resolves #484

isoutils/isoutils.go
```diff
@@ -123,8 +123,8 @@ func mkInitrd(version string, model *model.SystemInstall, options args.Args) err
 	options.SwupdFormat = "staging"
 	sw := swupd.New(tmpPaths[clrInitrd], options)
 
-	/* Install os-core only as initrd */
-	if err := sw.VerifyWithBundles(version, model.SwupdMirror, []string{}); err != nil {
+	/* Install os-core and os-core-plus as initrd */
```
mdhorn:
```suggestion
	/* Install os-core and os-core-plus (we only need kmod-bin) as initrd */
```

miguelinux:
done

mdhorn:
@miguelinux Verified that new images built with this change booted. Thank you!

PR #500, opened 2019-09-09
Disable AES checks.

ahkok:
We're well into disabling AES as a requirement at this point and have done the needed recompiles. At this point we're needlessly slowing down for little use - AES issues are quickly fixed if found.


puneetse:
Reference: https://github.com/clearlinux/distribution/issues/785

Issue #234, opened 2019-02-21
Please add support for offline installation


thiagomacieira:
**Is your feature request related to a problem? Please describe.**
Sometimes, one doesn't have network access on the machine Clear is being installed on. Being able to perform an offline installation, even if partial, would be very welcome.

**Describe the solution you'd like**
Add a way to install Clear onto the new machine without requiring Internet access. Suggestion: simply copy the installer system's /usr to the destination, then use swupd to remove the installer bundle and add the rest.

An added bonus for the graphical installer is that it would skip the need to download the entire desktop again. Only the deltas from the installer's set up to the current Clear would be necessary (plus of course any bundles the user chose to add during installation).

See also #233, which would argue for copying the whole installer to the new partitions.

**Describe alternatives you've considered**
Having network access.

**Additional context**

ajipw:
Yes, please provide an offline installation, my internet connection is slow. I think downloading an ISO of 2.17GB is useless because it has to be downloaded again when installing (with the same size)

BruceCherniak:
Offline Installation is **CRITICAL**, not just *nice to have* !!!

I envision an entire corporate market behind firewalls/proxies that make the current install difficult/impossible.  Our current wired network is locked down with 802.1x, preventing initial install.  Once installed there are methods to add the device to the network, but in its current state Clear Linux is completely unusable because it will not install without an open network.

I repeat, **in its current state, Clear Linux is completely unusable**

The live ISO *should* contain everything necessary for the base install.  After the initial install and network configuration, then it can check for updates.  Easy.  Every major linux distro provides this ability.

Why make this so *unnecessarily* difficult?!

mdhorn:
Thanks for your feedback.

We have a backlog of items to work on, and currently, the offline install is not at the top of our list.

This is an open source project, and we are happy to allocate time to reviewing pull request submissions.

BruceCherniak:
Hi Mark, thanks for your quick response.  I'm happy to lend a hand.  But, the first thing I notice about building the installer is the CLR-INSTALLER dependency (``swupd bundle-add clr-installer``).  Chicken, meet egg.  :-)

Luraktinus:
I really need an offline install too, would be cool to be able to download bundles/packs and import them too, i have customers without internet and i would like to provide them with software and updates by giving them a USB drive

jwakre:
[...]

PR #504, opened 2019-09-09
Add offline install support

jwakre:
Fixes Issue: #234 

- Add offline install support
- TUI: Update network check workflow for offline installs
- GUI: Update network check workflow for offline installs
- GUI: Add network check menu entry
- GUI: Create pop-up box for failing system checks
- GUI: Remove pre-check page


controller/controller.go
```diff
@@ -769,3 +778,30 @@ func generateISO(rootDir string, md *model.SystemInstall, options args.Args) err
 
 	return err
 }
+
+// writeCustomConfig creates a config file in the installer image that overrides the default
+func writeCustomConfig(chrootPath string, md *model.SystemInstall) error {
+	customPath := path.Join(chrootPath, conf.CustomConfigDir)
+
+	info, err := os.Stat(customPath)
+	if os.IsNotExist(err) {
```
dorileo: Why not use `utils.FileExists()` ? Also I don't think you need to check if it's directory since the environment is very controlled, we're pretty sure there won't be a directory in that path.

swupd/swupd.go
```diff
@@ -378,6 +411,29 @@ func (s *SoftwareUpdater) VerifyWithBundles(version string, mirror string, bundl
...
+	s.stateDir = filepath.Join(s.rootDir, conf.OfflineContentDir)
+
+	// Set swupd's --path argument to an empty directory so that all listed bundles will be
+	// downloaded (Swupd skips over previously installed bundles)
+	if s.rootDir, err = ioutil.TempDir("", "installerTmp-"); err != nil {
+		return err
+	}
+	defer func() { _ = os.RemoveAll(s.rootDir) }()
+
+	return s.VerifyWithBundles(version, mirror, bundles)
```
mdhorn: Now, if you download offline content and are generating an ISO, we call `VerifyWithBundles` three times -- the output is nearly identical each time and could be misleading to the user.

Please add a new parameter to  `VerifyWithBundles` with a prefix string for all of the steps. For the main OS install we can use a prefix like 'Target OS', for the offline content 'Offline Content', and for the ISO, 'ISO Boot OS'.

nazar2sfive:
what is the progress? 
without this we have to download content twice with 256KiB/s.

jwakre:
I made the following changes:
* Addressed review feedback
* Added documentation
* Improved offline install performance 
* Fixed indeterminate progress bar

controller/controller.go
```diff
@@ -581,6 +593,41 @@ func contentInstall(rootDir string, version string, md *model.SystemInstall, opt
 	return nil, nil
 }
 
+func copyOfflineToStatedir(rootDir, stateDir string) error {
```
jwakre: This function doesn't handle the case when the statedir already exists.

scripts/InstallerYAMLSyntax.md
```diff
@@ -117,6 +123,7 @@ Item | Description | Default
 `hostname` | Name of the host system | `-UNIQUE RANDOM-`
 `version` | Version of Clear Linux OS to install | `-VERSION_ON_BUILD_SYSTEM-`
 `autoUpdate` | Should the system automatically update to the latest release of Clear Linux OS as part of the installation?; true or false | true
+`offline` | Install update content for minimal offline installation | false
```
jwakre: I'm changing offline to default true.

isoutils/isoutils.go
```diff
@@ -112,6 +113,46 @@ func mkRootfs() error {
 	return err
 }
 
+// GetIsoLoopDevice returns the name of the ISO loop device
+func GetIsoLoopDevice() string {
+	w := bytes.NewBuffer(nil)
+	args := []string{
+		"losetup",
+		"-j",
+		"/mnt/media/images/roootfs.img",
```
jwakre: Typo:

s/roootfs/rootfs

jwakre:
@mdhorn I found 3 more issues that I will push fixes for soon:

1. Misspelled `/mnt/media/images/rootfs.img` which prevented ISO from using unsquashfs
2. When toggling Gparted with advanced disk partitioning, the welcome page is loaded instead of the menu page
3. When a system check fails in the GUI, 2 dialog windows are created

jwakre:
I made the following changes:

1. Fixed typo (roootfs -> rootfs) which re-enables support for unsquashfs
2. Updated the welcome page logic to properly reload the clr-installer when switching from Gparted and to prevent 2 dialog pop-ups when the system check fails.
3. Increased the indeterminate progress bar speed.

mdhorn:
LGTM 👍

nazar2sfive:
version : clear-31090-live-desktop.iso

it fails to proceed if internet is not detected.
image

@mdhorn from which version is offline install supported?

jwakre:
@nazar2sfive The installer images at https://clearlinux.org/downloads do not have offline install support yet. The clr-installer must make a release before images with offline installer support can be distributed which should happen soon. I will close #234 when installer images with offline support are published.

malventano:
Did this capability regress? Attempting a basic install of `clear-41190-live-server.iso` without network still attempts to connect to the network and fails. No packages were selected. Full defaults.

Issue #506, opened 2019-09-10
Replace locales with locales with glibc-locale


mdhorn:
The Clear Linux* OS bundle `locales` is a duplicate of `glibc-locale`.

**Describe the solution you'd like**
Replace the use of `locales` with `glibc-locale` in the code
https://github.com/clearlinux/clr-installer/blob/d10becc59715302b654f70ecd43d957686d702e9/language/language.go#L34
and in the clr-installer bundle definition.

**Describe alternatives you've considered**
Replacing the references to `glibc-locale` in all other bundle definitions since `locales` was created first. However, `glibc-locale` is a more correct bundle name.

mdhorn:
Released in clr-installer version 2.2.0 in Clear Linux* OS release 31030.

PR #509, opened 2019-09-11
locale: switch to swupd bundle glibc-locale

mdhorn:
Fixes Issue: #506

- Updating the installer to use the preferred swupd bundle glibc-locale instead of locales. Both have the same definition.


dorileo:
looks good to me

Issue #515, opened 2019-09-13
Graphical Installer does not choose the previously selected disk


mdhorn:
**Describe the bug**
After selecting an installation disk target in the graphical installer, returning to the "Select Installation Media" screen, the first disk in the list is always selected -- not necessarily the one which was previously selected. 
If the user selects the Confirm button they may be selecting a different disk than intended if they are not paying attention.

**To Reproduce**
1. Go to "Select Installation Media" screen
2. Select a disk from the pull-down list that is not the first one
3. Click CONFIRM
4. Go to "Select Installation Media" screen again
5. Note that the selected media is the first disk in the list!

**Expected behavior**
The previously selected disk should be the default when returning to the screen.

**Environment (please complete the following information):**
 - Clear Linux OS Version: 2.1.1
 
**Additional context**
You will need to ensure there is more than one possible disk for use in the installation; larger >20GB USB disk can be used.

mdhorn:
Fixed in clr-installer release 2.3.0 available in Clear Linux OS 31150 or greater.

Issue #521, opened 2019-09-17
Confirm button should be disabled when Destructive Installation options exist


mdhorn:
**Describe the bug**
The "CONFIRM" button should be disabled if there are no possible Destructive Installation options available in the media pull-down list.

**To Reproduce**
1. Ensure that only media connected are less than < 20MiB
2. Launch the graphical installer (clr-installer-gui)
3. Click through to the Select Installation Media screen
4. Select the "Destructive Installation" radio button.
5. The user should an empty choose list, be presented with a warning message about no possible choices (similar to the Safe Installation warning), and the "CONFIRM" button should be disabled.

**Expected behavior**
A Warning should be presented and the "CONFIRM" button disabled when there are no selectable Destructive options.

**Environment (please complete the following information):**
 - Clear Linux OS Version: [swupd info]
Distribution:      Clear Linux OS
Installed version: 31030
 - Installer Version [clr-installer -v]
2.2.0
 
**Additional context**
The latest release of the clr-installer is supposed to screen out media which is too small.

mdhorn:
Fixed in clr-installer release 2.3.0 available in Clear Linux OS 31150 or greater.

Issue #520, opened 2019-09-17
Safe Installation should exclude media that is too small


mdhorn:
**Describe the bug**
The Safe Installation option should filter out media that is smaller than the required minimum disk size for a Desktop Installation of 20MiB.

**To Reproduce**
1. Connect a blank USB < 20MiB
2. Launch the graphical installer (clr-installer-gui)
3. Click through to the Select Installation Media screen
4. The user should not see the media less than 20MiB

**Expected behavior**
Insufficiently sized media is not shown as an option

**Environment (please complete the following information):**
 - Clear Linux OS Version: [swupd info]
Distribution:      Clear Linux OS
Installed version: 31030
 - Installer Version [clr-installer -v]
2.2.0
 
**Additional context**
The latest release of the clr-installer is supposed to screen out media which is too small.

mdhorn:
Fixed in clr-installer release 2.3.0 available in Clear Linux OS 31150 or greater.

PR #523, opened 2019-09-17
Media selection bug fixes

mdhorn:
Fixes Issue: #515 #520 #521

- Fix default disk selection bug in the graphical interface
- Handle the case when there are no media with enough capacity for installing
- Ensure that the Safe media has enough capacity for an install


storage/ops.go
```diff
@@ -1334,7 +1334,7 @@ func FindSafeInstallTargets(rootSize uint64, medias []*BlockDevice) []InstallTar
 			continue
 		}
 
-		if curr.Children == nil || len(curr.Children) == 0 {
+		if (curr.Children == nil || len(curr.Children) == 0) && curr.Size > minSize {
```
jwakre: >=?

mdhorn: Good catch

mdhorn:
@jwakre I've fixed your comment. Please re-review when you have time.

gui/pages/disk_config.go
```diff
@@ -784,18 +784,22 @@ func (disk *DiskConfig) populateComboBoxes() error {
...
 					log.Warning("SetValue safeStore")
 					return err
 				}
+				if target.Name == disk.model.InstallSelected[target.Name].Name {
+					selected = n
+				}
 			}
 			disk.chooserCombo.SetModel(safeStore)
-			disk.chooserCombo.SetActive(0)
+			disk.chooserCombo.SetActive(selected)
```
jwakre: Sometimes when I switch between `Safe Install` and `Destructive Install` my selection is not remembered

jwakre: will be done in a separate PR

PR #526, opened 2019-09-17
telemetry: execute additional telemetry steps

alexjch:
This change executes opt-in when users agree to provide crash information
to the clear linux team and opt-out when users do not agree.

Signed-off-by: Alex Jaramillo <alex.v.jaramillo@intel.com>


mdhorn:
Fixed in clr-installer release 2.3.0 available in Clear Linux OS 31150 or greater.

PR #529, opened 2019-09-18
telemetry: do not fail if bundle is not installed

alexjch:
This change removes the need for telemetrics bundle to be present for
the installer to work.

Signed-off-by: Alex Jaramillo <alex.v.jaramillo@intel.com>

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -751,18 +751,22 @@ func saveInstallResults(rootDir string, md *model.SystemInstall) error {
 	}
 
 	if md.IsTelemetryEnabled() {
-		log.Info("Copying telemetry records to target system.")
-		if err := md.Telemetry.CopyTelemetryRecords(rootDir); err != nil {
-			log.Warning("Failed to copy image Telemetry data")
-			errMsgs = append(errMsgs, "Failed to copy image Telemetry data")
-		}
-		log.Info("Running telemctl opt-in")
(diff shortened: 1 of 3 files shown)


controller/controller.go
```diff
@@ -750,7 +750,7 @@ func saveInstallResults(rootDir string, md *model.SystemInstall) error {
 		log.Info("Skipping archiving of Installation results")
 	}
 
-	if md.IsTelemetryEnabled() {
+	if md.IsTelemetryEnabled() && md.IsTelemetryInstalled() {
```
mdhorn: Please change this to have the IsTelemetryInstalled nested inside of the Enabled
```suggestion
	if md.IsTelemetryEnabled()  {
	    if md.IsTelemetryInstalled() {
```

controller/controller.go
```diff
@@ -764,6 +764,8 @@ func saveInstallResults(rootDir string, md *model.SystemInstall) error {
 		if len(errMsgs) > 0 {
 			return errors.Errorf("%s", strings.Join(errMsgs, ";"))
 		}
+	} else if md.IsTelemetryEnabled() && md.IsTelemetryInstalled() == false {
```
mdhorn: Then this becomes the else half of the IsTelemeteryInstalled
```suggestion
	} else  {
```

Issue #539, opened 2019-09-24
lint-core makefile rule outputs many errors


jwakre:
**Describe the bug**
An unreasonably long list of false positive errors are displayed when running linter commands

**To Reproduce**
1. sudo make lint-core

**Expected behavior**
No error messages

**Environment (please complete the following information):**
 - Clear Linux OS Version: 31090
 - Installer Version 2.2.0

**Additional context**
Caused by this Makefile command: `go build -race github.com/clearlinux/clr-installer/...`

jwakre:
`go build` commands generate the warning dump, so this is not limited to the `lint-core` target

PR #545, opened 2019-09-25
Makefile: Suppress deprecation warnings

jwakre:
The gotk3 package contains deprecated packages that emit a large amount
of warnings during builds. Appending the -Wno-deprecated-declarations
flag to CGO_CFLAGS suppresses these warnings.

Fixes Issue: #539

- Add -Wno-deprecated-declarations option to CGO_CFLAGS in makefile which suppresses the very, very long list of gotk3 deprecated declaration warnings

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,15 @@ cov_dir = $(top_srcdir)/.coverage
 
 nproc = $(shell nproc)
 orig_go_path = $(shell go env GOPATH)
+
+# The -Wno-deprecated-declarations option is required to suppress deprecation
+# warnings in the gotk3 package. https://github.com/gotk3/gotk3/issues/420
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,6 +12,15 @@ cov_dir = $(top_srcdir)/.coverage
 
 nproc = $(shell nproc)
 orig_go_path = $(shell go env GOPATH)
+
+# The -Wno-deprecated-declarations option is required to suppress deprecation
+# warnings in the gotk3 package.
```
mdhorn: Can we add a reference to https://github.com/gotk3/gotk3/issues/420
so that when it is finally fixed we can remove this work-around.

jwakre:
@mdhorn Thanks for the review, changes are addressed.

Issue #524, opened 2019-09-17
clr-installer --config to support HTTP (like kickstart)


seanhig:
**Is your feature request related to a problem? Please describe.**

Attempting to use clear linux in Packer where I would normally use CentOS or RHEL.

**Describe the solution you'd like**

Have the clr-installer support passing an HTTP url to the --config parameter such that the same mechanism used by RHEL kickstart can be used with Packer (where the live-server.yaml configuration can be delivered over HTTP to the bootstrap).  This simplifies the installation process and enables existing Packer tooling to be used to create clear linux system images.

Eg.

clr-installer --config http://packerhost:8000/live-server.yaml

**Describe alternatives you've considered**

N/A

mdhorn:
@seanhig Thanks for your request.
This should not be too difficult as we already support using a URL for the configuration file when supplied as a kernel boot argument.

mdhorn:
Fixed in clr-installer release [2.3.1](https://github.com/clearlinux/clr-installer/releases/tag/2.3.1) in Clear Linux OS release 31270 or later.

seanhig:
Fantastic everyone, thank you so much.  I will test this ASAP.

PR #546, opened 2019-09-26
Add support to download config over network

jwakre:
When passing a config file URL to the --config flag, the installer will
download and consume the config. For example:

clr-installer --config https://pathToConfig.yaml

Signed-off-by: John Akre <john.w.akre@intel.com>

Fixes Issue: #524

- Support URL arguments to --config flag
- Move IsValidMirror to the network package and rename ot IsValidUrl

### DIFF
--- a/clr-installer/main.go
+++ b/clr-installer/main.go
@@ -222,7 +222,16 @@ func execute(options args.Args) error {
 		if cf, err = conf.LookupDefaultConfig(); err != nil {
 			return err
 		}
+	} else if network.IsValidURI(options.ConfigFile, options.AllowInsecureHTTP) {
+		if cf, err = network.FetchRemoteConfigFile(options.ConfigFile); err != nil {
+			fmt.Printf("Cannot acesss configuration file %q: %s\n", options.ConfigFile, err)
+			return err
+		}
+		options.CfDownloaded = true
+	} else {
+		return errors.Errorf("No valid configuration file")
 	}
+
 	if options.CfDownloaded {
 		defer func() { _ = os.Remove(cf) }()
 	}
--- a/clr-installer/main.go
+++ b/clr-installer/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/clearlinux/clr-installer/language"
 	"github.com/clearlinux/clr-installer/log"
 	"github.com/clearlinux/clr-installer/model"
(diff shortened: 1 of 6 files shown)


jwakre:
@mdhorn Currently, this doesn't work when the config file relies on pre/post install scripts because the scripts will not be downloaded. Although, this is likely an existing issue for the kernel boot argument.

clr-installer/main.go
```diff
@@ -216,7 +217,13 @@ func main() {
 		if cf, err = conf.LookupDefaultConfig(); err != nil {
 			fatal(err)
 		}
+	} else if network.IsValidUrl(options.ConfigFile) {
+		if cf, err = network.FetchRemoteConfigFile(options.ConfigFile); err != nil {
+			fatal(err)
```
jwakre: I'll improve error handling here.

clr-installer/main.go
```diff
@@ -221,7 +222,13 @@ func execute(options args.Args) error {
 		if cf, err = conf.LookupDefaultConfig(); err != nil {
 			return err
 		}
+	} else if network.IsValidURI(options.ConfigFile, options.AllowInsecureHTTP) {
+		if cf, err = network.FetchRemoteConfigFile(options.ConfigFile); err != nil {
+			return err
+		}
+		options.CfDownloaded = true
 	}
```
mdhorn: We need to add an else here since we do not have a valid configuration file.
```suggestion
	} else {
	    return errors.Errorf("No valid configuration file")	    
	}
```

clr-installer/main.go
```diff
@@ -221,7 +222,13 @@ func execute(options args.Args) error {
 		if cf, err = conf.LookupDefaultConfig(); err != nil {
 			return err
 		}
+	} else if network.IsValidURI(options.ConfigFile, options.AllowInsecureHTTP) {
+		if cf, err = network.FetchRemoteConfigFile(options.ConfigFile); err != nil {
+			return err
```
mdhorn: We need to output a message to the user here otherwise the just get an exit status number.
```suggestion
			fmt.Printf("Cannot acesss configuration file %q : %s\n", options.ConfigFile, err)
			return err
```

jwakre:
@mdhorn I added the additional feedback that you suggested.

Issue #478, opened 2019-07-19
Support insecure HTTP mirror


jwakre:
**Describe the solution you'd like**
HTTP mirror support should be enabled when the user acknowledges that the install may be insecure. HTTP support must enabled with an extra step, like a checkbox or flag that notifies the user that the mirror may be insecure.

reaganlo:
Instead of a checkbox it could be a popup message dialog with two buttons (Yes, No).

mdhorn:
Besides the User Interface for acknowledging the use of HTTP, we need to add a command-line argument so that one can build images using an HTTP URL. This would be need for such things a building from a local developer mix.

Consider using the same flag as swupd
`--allow-insecure-http`     `Allow updates over insecure connections`

mdhorn:
Once an insecure connection is confirmed, we should create a local swupd configuration file to retain the option:
```
cat /etc/swupd/config
[GLOBAL]
# Allow updates over insecure connections
allow_insecure_http=true
```

To ensure this propagates to the target system, the installer should copy the contents of /etc/swupd to the target system similar to what is done for the network settings.

PR #552, opened 2019-09-28
Add Support for insecure updates over HTTP

mdhorn:
Fixes Issue: #478

- Support insecure updates over HTTP
- swupd: identify certificate errors
- Ensure we default to software auto update for UI

### DIFF
--- a/clr-installer/main.go
+++ b/clr-installer/main.go
@@ -1,4 +1,4 @@
-// Copyright © 2018 Intel Corporation
+// Copyright © 2019 Intel Corporation
 //
 // SPDX-License-Identifier: GPL-3.0-only
 
@@ -131,10 +131,6 @@ func execute(options args.Args) error {
 
 	log.Info(path.Base(os.Args[0]) + ": " + model.Version +
 		", built on " + model.BuildDate)
-
-	if options.SwupdContentURL != "" && swupd.IsValidMirror(options.SwupdContentURL) == false {
-		return errors.Errorf("swupd-contenturl %s must use HTTPS or FILE protocol", options.SwupdContentURL)
-	}
 
 	if options.PamSalt != "" {
 		hashed, errHash := encrypt.Crypt(options.PamSalt)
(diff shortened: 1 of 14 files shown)


clr-installer/main.go
```diff
@@ -270,6 +266,14 @@ func main() {
 		md.SwupdMirror = options.SwupdMirror
 	}
 
+	if options.AllowInsecureHTTPSet {
+		md.AllowInsecureHTTP = options.AllowInsecureHTTP
+	}
+
+	if options.SwupdContentURL != "" && swupd.IsValidMirror(options.SwupdContentURL, md.AllowInsecureHTTP) == false {
+		fatal(errors.Errorf("swupd-contenturl %s must use HTTPS or FILE protocol", options.SwupdContentURL))
```
jwakre: With rebase this becomes return ....

swupd/swupd.go
```diff
@@ -539,6 +539,9 @@ func setMirror(swupdArgs []string, t string) (string, error) {
 // SetHostMirror executes the "swupd mirror" to set the Host's mirror
 func SetHostMirror(url string, allowInsecureHTTP bool) (string, error) {
 	if urlErr := network.CheckURL(url); urlErr != nil {
+		if strings.Contains(urlErr.Error(), "60") {
+			return "", fmt.Errorf(utils.Locale.Get("SSL certificate problem"))
```
jwakre: Use `errors.Errorf` to enable stack trace.

jwakre:
> When I try using an HTTP URL in the GUI, I get an error. Steps to reproduce
> 
>     1. Launch GUI
> 
>     2. Go to Software Updater Configuration
> 
>     3. Check `allow installation over insecure...`
> 
>     4. Enter the following URL: "http://cdn.download.clearliunx.org/update/"
> 
>     5. Hit Confirm
> 
> 
> I get the following error: Server does not report any verison

This URL seems to be problematic for curl. I'll try another one.

swupd/swupd.go
mdhorn: In this instance, we do not want the stack trace as this string is display to the user in the UI.

PR #554, opened 2019-09-29
"Failed to find installer media" error: add more details

stokito:
I had a problem when I created a USB drive with Unetbootin and it doesn't set a proper label `CLR_ISO` and I was really confused when I saw the `Failed to find installer media` message because I wasn't able to figure out what went wrong.
The PR adds this small but important detail to error message.
Also I created a corresponding ticket in the Unetbootin to add this label for ClearLinux https://github.com/unetbootin/unetbootin/issues/234

### DIFF
--- a/storage/ops.go
+++ b/storage/ops.go
@@ -1553,9 +1553,28 @@ func FormatInstallPortion(target InstallTarget) string {
 // ByBDName implements sort.Interface for []*BlockDevice based on the Name field.
 type ByBDName []*BlockDevice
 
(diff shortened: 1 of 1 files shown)


mdhorn:
@stokito Yes, please shorten the commit header and most the details to the body.
Along the lines of https://github.com/erlang/otp/wiki/writing-good-commit-messages

mdhorn:
@stokito The fact that the tool you used somehow dropped the required label concerns me that there may be more problems in using that tool. Just be aware, there may be more problems hiding there.

PR #555, opened 2019-09-30
scalable installer window, scrollable content

tpalli:
- attempt to use monitor relative size to the installer window if primary monitor found
- pack window.contentLayout to a scrolledwindow to ensure user is able to complete installation with low resolution monitor

### DIFF
--- a/gui/window.go
+++ b/gui/window.go
@@ -44,6 +44,11 @@ type Window struct {
 	contentLayout *gtk.Box    // Content Layout
 	rootStack     *gtk.Stack  // Root-level stack
 
+	scroller  *gtk.ScrolledWindow
+	useScroll bool // need to embed content layout to scrolled window
+
+	contentContainer *gtk.Container // main container for content
+
 	model   *model.SystemInstall // model
 	options args.Args            // installer args
 	rootDir string               // root directory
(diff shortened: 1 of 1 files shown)


gui/window.go
```diff
@@ -106,6 +109,33 @@ func (window *Window) CreateHeaderBar() error {
...
+	}
+
+	mon, err := dpy.GetPrimaryMonitor()
+	if err != nil {
+		return err
+	}
+
+	rect := mon.GetGeometry()
+
+	win.handle.SetDefaultSize(
```
mdhorn: I like that this will scale for smaller resolutions, but I think our current size is about the largest we should make the install screen. Can we calculate the relative size first, and then set the ceiling at our current defaults?
    


```suggestion
        relW := int(w*float32(rect.GetWidth())
        relH := int(h*float32(rect.GetHeight())
        if relW > WindowWidth || relH > WindowHeight {
             window.handle.SetDefaultSize(WindowWidth, WindowHeight)
        } else {
             window.handle.SetDefaultSize(relW, relH)
        }     
```

tpalli: Sure, that sounds fine approach to me. Will fix and re-test.

tpalli: OK, what I ended up doing looks a bit different but should do the same thing. I'm comparing to the actual monitor size and if we don't fit then I use the relative size.

Issue #502, opened 2019-09-09
When producing an image from a custom mix, "version: 0" wrongly points to upstream's latest _version


bktan8:
When producing an image from a custom mix, "version: 0" wrongly points to upstream's latest version instead of my custom mix's latest version number.  I have to specifically edit the YAML file each time the version is updated.  

Please change clr-installer to point </path/to/my_mix/update/www/version/latest_version>.

mdhorn:
Fixed in clr-installer release [2.3.1](https://github.com/clearlinux/clr-installer/releases/tag/2.3.1) in Clear Linux OS release 31270 or later.

PR #557, opened 2019-10-01
Update to latest when version is 0

jwakre:
When the version field in the config file is set to 0, the latest
version will be installed instead of the host's version.

Fixes Issue: #502

- The default version is the latest mix version. Previously, the default version was the host's version.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -94,7 +94,7 @@ func Install(rootDir string, model *model.SystemInstall, options args.Args) erro
 	}
 
 	if model.Version == 0 {
-		version = utils.ClearVersion
+		version = "latest"
 	} else {
 		version = fmt.Sprintf("%d", model.Version)
 	}
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -493,6 +493,7 @@ func contentInstall(rootDir string, version string, md *model.SystemInstall, opt
 		if err := utils.ParseOSClearVersion(); err != nil {
 			return prg, err
 		}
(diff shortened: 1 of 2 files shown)


jwakre:
> I'd like to hold this for a bit.
> 
> After talking with Dorileo we recalled the reason for using the current OS's version and not latest -- format bumps.

Good catch! We should investigate how format bumps are handled.
 
> The current version of swupd very likely can not generate an os-install for a newer format than itself.
> 
> The safest approach is to only allow os-installs for the current or older version of the OS (or mix). Or maybe the latest version for the same format.

I would expect `swupd os-install` to update to the latest version in the format. Otherwise, this is a bug. I will look into the current behavior of `swupd os-install`.

Downstream mixes are more complicated and there isn't much that we can do from the installer to always update to the expected latest version. Since downstream mixes can bump their format ahead of upstream Clear, building images based on a downstream mix with a host based on upstream Clear could lead to unexpected results. Mixer/swupd would need to identify their compatibility in another way that is not as easily changed as the format number which can be altered for other reasons, like deleting bundles.

jwakre:
@mdhorn I confirmed that the current behavior of `swupd os-install --version latest ....` is to install the latest version for the host's format. To enable downstream mixes that have formats ahead of the host system, we can allow users to set swupd's `--format` argument. Below is an example to create a chroot with the latest version in format 50:

`swupd os-install --version latest --format 50 ......`

I will open a separate issue for that.

jwakre:
@mdhorn Added a log message to capture offline install version overrides.

PR #561, opened 2019-10-03
Config bugfix

jwakre:
- Fix bug where relative/absolute config file paths were rejected
- Add HTTP warning when using HTTP URI without seeing --allow-insecure-http

### DIFF
--- a/clr-installer/main.go
+++ b/clr-installer/main.go
@@ -228,8 +228,8 @@ func execute(options args.Args) error {
 			return err
 		}
 		options.CfDownloaded = true
-	} else {
-		return errors.Errorf("No valid configuration file")
+	} else if ok, err := utils.FileExists(options.ConfigFile); !ok || err != nil {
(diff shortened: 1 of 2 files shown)


clr-installer/main.go
```diff
@@ -228,7 +228,7 @@ func execute(options args.Args) error {
 			return err
 		}
 		options.CfDownloaded = true
-	} else {
+	} else if ok, err := utils.FileExists(options.ConfigFile); !ok || err != nil {
 		return errors.Errorf("No valid configuration file")
```
mdhorn: Perhaps we should use the same or similar messages the file local file is not there?
```suggestion
		return errors.Errorf("Cannot acesss configuration file %q: %s", options.ConfigFile, err)
```

jwakre: I can do:
`return errors.Errorf("Cannot acesss configuration file %q", options.ConfigFile)`

The error will typically be `nil` when the file doesn't exist. It seems strange to print that.

Issue #553, opened 2019-09-29
GUI Advanced installation: unable to find partitions to install CLR_BOOT, CLR_SWAP, CLR_ROOT


stokito:
I runned GParted and created all needed partitions but when I started the GUI installer again it still shows me "Missing CLR_BOOT partition label, Missing CLR_SWAP partition label, Missing CLR_ROOT partition label". Even when I pressed the Rescan media button the error persists.

But when I runned TUI installer in terminal `clr-installer` it was able to find this partitions and installation was successful.
See screenshots for details.
![GUI](https://user-images.githubusercontent.com/415502/65824266-89974300-e255-11e9-947e-641a08f620c4.png)

![TUI](https://user-images.githubusercontent.com/415502/65824264-89974300-e255-11e9-9cb6-7a9e54d25598.png)

![GParted](https://user-images.githubusercontent.com/415502/65824265-89974300-e255-11e9-9744-a00417d6338c.png)

**Environment (please complete the following information):**
 - Clear Linux OS Version: 31090
 - Installer Version 2.2.0

mdhorn:
@stokito Were any of the partitions from /dev/sda mounted on the running OS? If so, we will mask that device as a possible install target to prevent modifying media which is in use.

The default ISO image should be disabling automount of media for this reason.

mdhorn:
If you have the /root/clr-installer.log it may provide some insight as to why it was not seen.

stokito:
[clr-installer.log](https://github.com/clearlinux/clr-installer/files/3680223/clr-installer.log)
From what I see the installer failed because of the `/dev/loop0` device.
I executed the command  `lsblk --exclude 1,2,11 -J -b -O`:
```json
{
   "blockdevices": [
[...]

PR #562, opened 2019-10-03
gui: Correctly set advanced disk config warnings

jwakre:
When using advanced disk partitioning with an invalid setup and loading
the menu page or disk configuration page for the first time, the warning
messages were not set correctly. The warnings were incorrectly set
because the TargetMedias variable was unexpectedly overwritten to nil.
This change prevents unexpected overwrites of TargetMedias and sets
TargetMedias when it is uninitialized.

Fixes: #553

- Correctly set menu disk config warning when incorrectly setting advanced disk partitioning
- Correctly set disk config warning when incorrectly setting advanced disk partitioning

### DIFF
--- a/gui/pages/disk_config.go
+++ b/gui/pages/disk_config.go
@@ -90,7 +90,13 @@ func (disk *DiskConfig) advancedButtonToggled() {
 	results := storage.DesktopValidateAdvancedPartitions(disk.model.TargetMedias)
 	if len(results) > 0 {
 		disk.model.ClearInstallSelected()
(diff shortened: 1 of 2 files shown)


jwakre:
still a little buggy

jwakre:
It seems like sometimes `storage.ListAvailableBlockDevices` doesn't find the block devices correctly..

jwakre:
@mdhorn Advanced disk warnings are more stable now with `udevadm settle`

PR #573, opened 2019-10-08
Swupd version via command line

mdhorn:
- Allow setting swupd version on the command line

### DIFF
--- a/clr-installer/main.go
+++ b/clr-installer/main.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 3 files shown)


clr-installer/main.go
```diff
@@ -289,6 +290,14 @@ func execute(options args.Args) error {
 	if options.SwupdFormat != "" {
 		md.SwupdFormat = options.SwupdFormat
 	}
+	if options.SwupdVersion != "" {
+		version, err := strconv.ParseUint(options.SwupdVersion, 10, 32)
```
jwakre: "latest" is a valid input for swupd's version argument. Can you add check for this case.

jwakre: You should set the version to 0 when latest is provided

mdhorn: @jwakre Good catch.  We only allow an unsigned in the YAML so I was following the same convention for the command line. But now allow `latest`.
Force pushed a fix.

Issue #544, opened 2019-09-25
Text-base UI should ensure root account is secured for the install target


bktan8:
Currently, when you install the server version, root's password is left unset through these scenarios:
1) If you just fulfill the "Required" settings, root is left unset and no user is added
2) If you add a user in the "Advanced" tab, unless the user is specifically set as administrator, the root password is left unset.

Scenario #2 is the one I'm more concerned about because you added a user, but you may forget to set a password for root.  Albeit a low probability, someone can connect to the system and hijack the root account.  

My proposal is to move the "add user a new user" screen to the "Required" tab (like the Desktop installer) and require that a user is added and the "Administrator" checkbox is set by default.  Secure the root account by default.  Don't leave it open.

mdhorn:
@bktan8 Historically, the text-based UI is for server installs which generally do not have user accounts.

The requirement should be:
- Set the root password
OR
- Create a user account with Administrator privileges

Yes, as a requirement it would move to the required screen.

jwakre:
The installer will require that a user account with admin privileges is created, but support will not be added to create a root user at install time.

mdhorn:
Fixed in https://github.com/clearlinux/clr-installer/releases/tag/2.3.2

mdhorn:
Available in Clear Linux OS release 31380 and higher.
https://clearlinux.org/downloads

PR #576, opened 2019-10-09
tui: Secure root user

jwakre:
Fixes Issue: #544

- Move manage user page to required tab in tui
- Add warning and disable confirm button when no admin user is created in tui
- Check the admin button by default for user creation in tui
- Improve logic for ensuring that all required pages are completed in tui

### DIFF
--- a/tui/user_manager.go
+++ b/tui/user_manager.go
@@ -58,6 +58,7 @@ func init() {
 type UserManagerPage struct {
 	BasePage
 	scrollingFrame *clui.Frame // content scrolling frame
+	warningLabel   *clui.Label // content warning label
 
 	columnFormat string
 
@@ -69,6 +70,16 @@ type UserManagerPage struct {
 	activeRow   *clui.Frame
 	activeUser  *SimpleButton
 	activeLogin string
+}
+
+// GetDone Returns true when there is an admin user
+func (page *UserManagerPage) GetDone() bool {
+	for _, user := range page.getModel().Users {
+		if user.Admin {
+			return true
+		}
+	}
+	return false
 }
 
 // GetConfiguredValue Returns the string representation of currently value set
@@ -152,7 +163,14 @@ func (page *UserManagerPage) Activate() {
 		page.scrollToRow(page.activeRow)
 	}
 
-	if len(page.users) > 0 || page.usersChanged {
+	adminSet := false
+	for _, user := range page.users {
+		if user.Admin {
+			adminSet = true
+		}
+	}
+
+	if adminSet {
 		page.confirmBtn.SetEnabled(true)
 	} else {
 		page.confirmBtn.SetEnabled(false)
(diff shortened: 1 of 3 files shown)


tui/user_manager.go
```diff
@@ -182,13 +199,29 @@ func (page *UserManagerPage) redrawRows() {
...
+				adminSet = true
+			}
 			if err := page.addUserRow(user); err != nil {
 				page.Panic(err)
 			}
 		}
+
+		// Display a warning when no admin user is set
+		if !adminSet {
+			rowFrame := clui.CreateFrame(page.scrollingFrame, 1, 1, clui.BorderNone, clui.Fixed)
```
mdhorn: The warning message should be below the scrolling frame, not a part of it.
If you attempt to add 6 accounts, all non-admin, the warning will be hidden and not scrollable.

jwakre: Good catch, I'll fix this

jwakre: @mdhorn The scrollable frame is causing me issues. I created a new frame below the scrollable frame and placed the warning label inside. The scrollable frame pushes the newly created frame off of the screen. Even when I force the scrollable frame to use a small amount of space, the hidden (scrollable) elements still push the warning label lower... Any good ideas?

mdhorn: @jwakre Make the scrollable frame one line smaller when it is created in `newUserManagerPage`.
Created the new frame for the warning label also in the `newUserManagerPage` function, and it should be added to the `page.content`, not the scrollable frame.

`page.warningFrame = clui.CreateFrame(page.content, 1, 1, clui.BorderNone, clui.Fixed)`

jwakre: I tried that approach. Even though there is plenty of screen real estate below the scrollable frame, as I add elements to the scrollable frame, the warning label in the newly created, page.content child frame is still pushed lower. It feels like a clui bug to me.

mdhorn: @jwakre Yep, looks like a Clui bug.
I have an idea, give me a few minutes to try sometime.

jwakre:
@mdhorn Thanks for the patch. I squashed it and looks good.

mdhorn:
Drop commit 5acb855b8196d03d0014b2140ed32f9b984ffd3c
What specifically were you try to solve with that change?

jwakre:
> Drop commit [5acb855](https://github.com/clearlinux/clr-installer/commit/5acb855b8196d03d0014b2140ed32f9b984ffd3c)
> What specifically were you try to solve with that change?

This change prevents the the install button from becoming enabled until all required pages are "done". Previously, the install button was enabled when telemetry was "user defined" and the model was validated. This was problematic when I added the "manage user" page to the required tab.

It seems that the media page doesn't implement the GetDone() function as I expected. If GetDone() returns true for the media page in this case, things will work.

mdhorn:
https://github.com/clearlinux/clr-installer/pull/578 fixes the media page GetDone

mdhorn:
> > Drop commit [5acb855](https://github.com/clearlinux/clr-installer/commit/5acb855b8196d03d0014b2140ed32f9b984ffd3c)
> > What specifically were you try to solve with that change?
> 
> If we drop that commit, the "manage users" page will still be optional.
> 
> This change prevents the the install button from becoming enabled until all required pages are "done". Previously, the install button was enabled when telemetry was "user defined" and the model was validated. This was problematic when I added the "manage user" page to the required tab.
> 
> It seems that the media page doesn't implement the GetDone() function as I expected. If GetDone() returns true for the media page in this case, things will work.

Okay, looks like we just need to also have PR #578 with you current code for things to work.

PR #578, opened 2019-10-10
tui: Set done when media selected

mdhorn:
- Fixed bug where the media was not complete for advance disk configuration


tui/media_config.go
```diff
@@ -73,6 +74,7 @@ func (page *MediaConfigPage) GetConfiguredValue() string {
 		if storage.AdvancedPartitionsRequireEncryption(tm) && model.CryptPass == "" {
 			return fmt.Sprintf("Warning: %s", "Encryption passphrase required")
 		}
+
```
mdhorn: This should not be added

PR #585, opened 2019-10-14
Remove telemctl from travis

jwakre:
The docker image used by travis doesn't have telemctl by default any more and telemetry is optional for the installer. This change removes the call to telemctl.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,6 @@ before_script:
     - docker exec -it clear-test bash -c "swupd info"
       # Debug information
     - docker exec -it clear-test bash -c "set;printenv"
-      # Force restart telemetry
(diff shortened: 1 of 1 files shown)


jwakre:
https://travis-ci.com/clearlinux/clr-installer/builds/131866266 is an example failure where telemctl does not exist in the container.

mdhorn:
> https://travis-ci.com/clearlinux/clr-installer/builds/131866266 is an example failure where telemctl does not exist in the container.

Telemetry is now optional for the installer. There have been many updates about how telemetry is handled for the clr-installer. The better option would be to remove the `telemctl restart` lines from the `.travis.yml` file. That should work too.

jwakre:
@mdhorn Now the call to telemctl is removed.

PR #593, opened 2019-10-26
mixer: Add image template for use by mixer

mdhorn:
- Add template file for mixer to enable clr-installer as the build tool.


mdhorn:
> The problem of moving this into mixer is that we would need to monitor changes to the file format. In the future, maybe clr-installer could have a "command line template generator" that we could call from mixer to get the template instead.

Adding a `--generate-config` option would be a good add for the future. Mixer could call clr-installer with that option during `mixer init`.

PR #595, opened 2019-10-30
Remove kbd and glibc-locale dependencies

jwakre:
The validation errors for the keyboard and language settings have been
changed to warnings. This enables the kbd and glibc-locale bundles to
be set as optional in the clr-installer bundle definition.

When the kbd and glibc-locale bundles are not installed on the host
system, the tui/gui will not populate the list of alternate keyboards
and locales and the mass installer will display a warning that it was
unable to validate the keyboard or locale.

Fixes Issue: Reduces bundle definition requirements

- Remove kbd and glibc-locale bundle dependencies on host system. They will become also-adds in the bundle definition.
- Mass Installer: keyboard and locale validation failures will become warnings instead of fatal errors.
- TUI/GUI: The list of available keyboards and languages will be reduced (english only) when the kbd and glibc-locale bundles are not present on the host.

### DIFF
--- a/clr-installer/main.go
+++ b/clr-installer/main.go
@@ -349,20 +349,34 @@ func execute(options args.Args) error {
 		}
 	}
 
+	// The kbd bundle is an also-add and may be removed
 	if md.Keyboard != nil && !keyboard.IsValidKeyboard(md.Keyboard) {
-		return fmt.Errorf("Invalid Keyboard '%s'", md.Keyboard.Code)
+		msg := fmt.Sprintf("Warning: Host failed to validate keyboard: %s, using %s",
+			md.Keyboard.Code,
+			keyboard.DefaultKeyboard)
+
+		fmt.Println(msg)
+		log.Warning(msg)
+		md.Keyboard.Code = keyboard.DefaultKeyboard
 	}
 
 	if md.Timezone != nil && !timezone.IsValidTimezone(md.Timezone) {
 		return fmt.Errorf("Invalid Time Zone '%s'", md.Timezone.Code)
 	}
 
+	// The glibc-locale bundle is an also-add and may be removed
 	if md.Language != nil && !language.IsValidLanguage(md.Language) {
-		return fmt.Errorf("Invalid Language '%s'", md.Language.Code)
-	}
+		msg := fmt.Sprintf("Warning: Host failed to validate language: %s, using %s",
+			md.Language.Code,
+			language.DefaultLanguage)
 
-	// Set locale
-	utils.SetLocale(md.Language.Code)
+		fmt.Println(msg)
+		log.Warning(msg)
+		md.Language.Code = language.DefaultLanguage
+	} else {
+		// Set locale
+		utils.SetLocale(md.Language.Code)
+	}
 
 	// Run system check and exit
 	if options.SystemCheck {


jwakre:
@mdhorn This is the minimum amount of work to remove the kbd/glibc-locale dependencies. If it is important to continue fatal keyboard/language validations, functionality to query the host system's bundle list must be added.

Note: I tested this on the mass-installer and TUI without kbd, glibc-locale, tzdata. I have not yet tested the GUI, but it should work similarly to the TUI.

clr-installer/main.go
```diff
@@ -349,21 +349,27 @@ func execute(options args.Args) error {
 		}
 	}
 
+	// The kbd bundle is an also-add and may be removed
 	if md.Keyboard != nil && !keyboard.IsValidKeyboard(md.Keyboard) {
-		return fmt.Errorf("Invalid Keyboard '%s'", md.Keyboard.Code)
+		msg := fmt.Sprintf("Warning: Host failed to validate keyboard: %s", md.Keyboard.Code)
+		fmt.Println(msg)
```
mdhorn: What happens in the TUI/GUI when we print to the screen?

clr-installer/main.go
```diff
@@ -349,21 +349,27 @@ func execute(options args.Args) error {
...
 
 	if md.Timezone != nil && !timezone.IsValidTimezone(md.Timezone) {
 		return fmt.Errorf("Invalid Time Zone '%s'", md.Timezone.Code)
 	}
 
+	// The glibc-locale bundle is an also-add and may be removed
 	if md.Language != nil && !language.IsValidLanguage(md.Language) {
-		return fmt.Errorf("Invalid Language '%s'", md.Language.Code)
+		msg := fmt.Sprintf("Warning: Host failed to validate language: %s", md.Language.Code)
+		fmt.Println(msg)
```
mdhorn: What happens in the TUI/GUI when we print to the screen?

clr-installer/main.go
jwakre: TUI: The warning is written to stdout on the terminal.
GUI: Untested

There are other instances of fmt.Println and fmt.Printf in this function, so it should behave similarly.,

mdhorn: Okay.

jwakre:
> Okay. The standard images would include all of the bundles so it should be fine.

So you are fine with **invalid keymaps and locales bypassing validation when using the mass installer**? For instance I can set the keymap to foo and the language to baz on the target system. Invalid settings seem to function identically to the default, but when querying the keyboard or language on the target system it will still show foo/bar.

mdhorn:
> > Okay. The standard images would include all of the bundles so it should be fine.
> 
> So you are fine with **invalid keymaps and locales bypassing validation when using the mass installer**? For instance I can set the keymap to foo and the language to bar on the target system. Invalid settings seem to function identically to the default, but when querying the keyboard or language on the target system it will still show foo/bar.
> 
> I just want to be clear that the validation will now be a warning instead of an error, even with all optional bundles. This is only a mass installer problem.

No, if the keyboard is invalid, we need to print the warning and then use the keyboard.DefaultKeyboard
Same for the language, warning and then use the built-in default of language.DefaultLanguage

jwakre:
@mdhorn Now when an invalid keyboard or language is used, a warning is displayed and the default value is used. When optional bundles are not installed, the installer will prevent alternate keyboard/language settings.

clr-installer/main.go
```diff
@@ -349,20 +349,34 @@ func execute(options args.Args) error {
...
-	}
+		msg := fmt.Sprintf("Warning: Host failed to validate language: %s, using %s",
+			md.Language.Code,
+			language.DefaultLanguage)
 
-	// Set locale
-	utils.SetLocale(md.Language.Code)
+		fmt.Println(msg)
+		log.Warning(msg)
+		md.Language.Code = language.DefaultLanguage
```
jwakre: Need to set md.Language.Tag and look at how important userDefined is.

jwakre:
I decided to drop this PR and leave the kbd bundle unchanged in the bundle definition. Without this bundle, keyboards cannot be easily validated.

Without making code changes, the tzdata bundle can be dropped and the glibc-locale bundle can be made optional. I have learned that the mass installer only support 3 languages, so the removal of glibc-locale does not have a huge impact.

PR #596, opened 2019-11-01
swupd: Support skipping optional bundles

jwakre:
Optionally included bundles (also-add) can now be omitted from the
install target by passing the --swupd-skip-optional flag or by setting
the swupdSkipOptional field in the config file.

- Add flag/config field to omit optional bundles from install target

### DIFF
--- a/args/args.go
+++ b/args/args.go
@@ -52,6 +52,8 @@ type Args struct {
 	AllowInsecureHTTP       bool
 	AllowInsecureHTTPSet    bool
 	CryptPassFile           string
+	SwupdSkipOptional       bool
+	SwupdSkipOptionalSet    bool
 	SwupdMirror             string
 	SwupdStateDir           string
 	SwupdCertPath           string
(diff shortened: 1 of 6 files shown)


args/args_test.go
```diff
@@ -434,6 +434,9 @@ func TestKernelAndCommandlineAllArgs(t *testing.T) {
 	if testArgs.SwupdMirror != "" {
 		t.Errorf("Command Line 'mirror' is not defaulted to ''")
 	}
+	if testArgs.SwupdSkipOptional != true {
```
mdhorn: This should default to false.

args/args.go
```diff
@@ -260,6 +262,10 @@ func (args *Args) setCommandLineArgs() (err error) {
 		&args.AllowInsecureHTTP, "allow-insecure-http", false, "Allow installation over insecure connections",
 	)
 
+	flag.BoolVar(
```
jwakre: I'm unlucky enough to go over the cyclomatic dependency limit...

Issue #591, opened 2019-10-18
Enable retries to download installation configuration file via kernel params


bktan8:
In some cases, we're seeing that attempt to download the installation template file fails.  My guess is that it's due to the network being slow to come up and thus an IP address is not available.  I would like to propose that we do 5 retries before failing the installation.

mdhorn:
@bktan8 This change needs to be made in the clr-installer as it is the application downloading the YAML configuration file.

mdhorn:
First released in 2.3.3, in Clear Linux OS 31530

bktan8:
Awesome work @mdhorn!  Thanks for reporting release version.

PR #597, opened 2019-11-04
PXE Enablement changes

mdhorn:
NOT READY FOR MERGE -- testing phase

Fixes Issue: #591

- Add a new service file for PXE installs
- Add retries to download the configuration file
- Properly enable automatic reboots

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ endif
 ifeq ($(IS_GIT_REPO),1)
 # Use the git tag and SHA
 # Standard build case from Git repo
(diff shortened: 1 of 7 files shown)


jwakre:
@mdhorn I still see that sometimes the service file is deleted before it runs.

mdhorn:
> @mdhorn I still see that sometimes the service file is deleted before it runs.

Are you using the tip of this PR?  I remove the disable Post command yesterday, so there should be nothing removing the link.

jwakre:
> Are you using the tip of this PR? I remove the disable Post command yesterday, so there should be nothing removing the link.

Ok, I probably made a mistake then. There are a lot of moving pieces in my test environment.

PR #601, opened 2019-11-14
mixer: generate template

mdhorn:
- Increase boot partition size in mixer template


model/model_ister.go
```diff
@@ -214,15 +215,29 @@ func JSONtoYAMLConfig(cf string) (string, error) {
...
+	// kernel modules by default: native or LTS
+	sort.Sort(sort.Reverse(sort.StringSlice(kernelBundles)))
+	var foundKernel bool
+	for _, bundle := range kernelBundles {
+		// Only use the first kernel bundle
+		if !foundKernel {
+			foundKernel = true
+			si.Kernel = &kernel.Kernel{Bundle: bundle}
+		} else {
+			fmt.Printf("Warning: Ignoring extra kernel bundle '%s' in YAML file; '%s' already in use.\n",
```
reaganlo: I have a PR in clr-installer, to skip the addtional kernel bundles if iso is set.
https://github.com/clearlinux/clr-installer/pull/604. If you feel this is a valid PR, can we avoid skipping the kernel bundles in mixer and let clr-installer handle it.

Issue #609, opened 2019-11-25
Not able create images on RAID drive


dlpottra:
**Describe the bug**
Not able to create install images with RAID drive

**To Reproduce**
1. sudo su -
2. swupd bundle-add clr-installer (if not installed)
3. clr-install -c cloudgues.yaml  (attached)

**Expected behavior**
should output a cloudguest.img file.  This works on a machine that doesn't have RAID drives configured.

**Environment (please complete the following information):**
 - Clear Linux OS Version: 31700
 - Installer Version 2.3.4

dlpottra:
[cloudguest.yaml.txt](https://github.com/clearlinux/clr-installer/files/3888524/cloudguest.yaml.txt)
[clr-installer.log](https://github.com/clearlinux/clr-installer/files/3888525/clr-installer.log)
[pre-install-clr-installer.yaml.txt](https://github.com/clearlinux/clr-installer/files/3888527/pre-install-clr-installer.yaml.txt)

mdhorn:
@dlpottra Can you attach the output of `lsblk --exclude 1,2,11 -J -b -O`?
We do not support using RAID in the installer, but if you can share the output I can make an update to ignore RAID devices instead of failing.

dlpottra:
[lsblk.txt](https://github.com/clearlinux/clr-installer/files/3889060/lsblk.txt)

FWIW I am trying to create images for a cloud.  I have RAID on my system that is creating the images.  I do know that the installer doesn't support RAID.  I don't think it should fail when trying to create images.

mdhorn:
Currently, the installer still parses all local media for possible install targets as that is a possible usage.
We can update it to be able to parse your raid drives even if they aren't valid targets.

mdhorn:
Fixed in clr-installer release 2.3.5.
Released in Clear Linux OS release 31750.

dlpottra:
Works good.  Thx.

PR #611, opened 2019-11-26
storage: Add RAID type to known list

mdhorn:
Fixes Issue: #609

- Parse RAID device types on host system; still not valid for install target


mdhorn:
> So it still does not support installing on RAID devices but does not prevent creating images on a host that has RAID devices. LGTM.

Correct. Similar to the changes we implemented to allow LVM volume on the host system where images are generated even though we still don't support LVM as an installation target.
Thanks for the review!

Issue #614, opened 2019-11-26
Hard Drive encryption changes the selected drive


maljaroudi:
**Describe the bug**
The hard drive encryption option changes the desired hard drive back to the first one on the list. Forgot to double check the first time and ended up wiping my secondary drive with personal files instead. 

**To Reproduce**
1. Go to 'Select Installation Media'
2. Choose 'Destructive Installation ' and select the second drive if you have two
3. Check "Enable Encryption"
4. Enter a passphrase and confirm
5. The drive selected will be switched back to the first on the list/ default

**Expected behavior**
The installer should never change the selected drive after the encryption is checked. Users might forget to double check that it changed back and end up losing their data.

**Environment (please complete the following information):**
 - Clear Linux OS Version: 31700
 - Installer Version : Latest ISO for version 31700
 - Platform: 2 Samsung SSDs. 1Tb, 256Gb.

mdhorn:
@maljaroudi Are you using the live-server or live-desktop? Was the problem seen in the text-based installer or the graphical version?

maljaroudi:
> @maljaroudi Are you using the live-server or live-desktop? Was the problem seen in the text-based installer or the graphical version?

Sorry i didn't include that in the report. 
live-desktop graphical version is what i used.

mdhorn:
@maljaroudi Thank you for the bug report. I was able to reproduce this.

mdhorn:
Fixed in Clear Installer release 2.3.7
Available in Clear Linux OS 31930 or later
https://clearlinux.org/downloads

PR #615, opened 2019-11-28
vendor: update all of the packages

mdhorn:
- Update to the latest version of each vendor package
  - Need a new gotk3 to fix a known bug


jwakre:
@mdhorn What is the known bug?

mdhorn:
> @mdhorn What is the known bug?

https://github.com/gotk3/gotk3/issues/361

mdhorn:
> > @mdhorn What is the known bug?
> 
> [gotk3/gotk3#361](https://github.com/gotk3/gotk3/issues/361)

I have a fix for #614 which is not pushed yet as it hit the bug noted above.

bryteise:
You may want to consider using https://dependabot.com/ at least for things that are versioned releases (oh if only all go things would version but I digress).

mdhorn:
> You may want to consider using https://dependabot.com/ at least for things that are versioned releases (oh if only all go things would version but I digress).

Sadly, 6 of 17 vendor packages we use do not have releases, and one of them has not bumped their release in 8 months though they continue to fix bugs. :cry: 

We have vendor code check, status, and update commands built into the Makefile. We update the vendor once a month when we do a security check or when we find a bug that is fixed.

PR #618, opened 2019-12-02
Fix GUI drive selection

mdhorn:
Fixes Issue: #614

- Ensure content is updated when rescan button is selected
- Retain disk selection when enabling encryption


gui/pages/disk_config.go
```diff
@@ -747,6 +756,30 @@ func (disk *DiskConfig) onEncryptClick(button *gtk.CheckButton) {
...
 
+func (disk *DiskConfig) onChooserComboChanged(combo *gtk.ComboBox) {
+	if iter, iterErr := combo.GetActiveIter(); iter != nil && iterErr == nil {
+		if model, modelErr := combo.GetModel(); modelErr == nil && model.GetNColumns() >= 2 {
+			// Extract from model
+			if valueObj, getErr := model.GetValue(iter, 2); getErr == nil && valueObj != nil {
+				if _, fType, typeErr := valueObj.Type(); typeErr == nil && fType == glib.TYPE_STRING {
+					if name, nameErr := valueObj.GetString(); nameErr == nil {
+						disk.tempSelectedTarget = name
+						log.Debug("ChooseBox entry selected is: %v", name)
```
reaganlo: Did you mean ChooseBox or ComboBox?

mdhorn:
> LGTM.
> 
> Though, there were few times during testing when the gui crashed - when I go out of the Media screen to the User screen and back to Media screen and hit Rescan. Though its not always reproducible and I'm not sure if its related to this PR.
> 
Hmmm..well, now I'll want to test this with the unmodified code. I thought we resolve these crash a while ago.  Did you save the log file from one of these crashes?

> Also, not related to the PR,
> If the Encryption is selected, and user clicks Advanced Installation and then clicks Destructive Installation again, the Encryption selection is not retained. Is it intended?
This is by design and how it functioned before this change.  We don't track and save settings per installation type. If you change the installation type, most things are reset to defaults.

mdhorn:
@reaganlo I'm no longer able to reproduce the crash. Can you retest on the system you had the failure?
I'll hold merging this until we know we haven't added a new bug while fixing the drive selection.

mdhorn:
@reaganlo I was finally able to generate a crash. I'm going to try backing gotk3 back to the sha of the fix I know I needed and test that since gotk3 hasn't made any official releases.
https://github.com/gotk3/gotk3/pull/366
https://github.com/gotk3/gotk3/commit/38a191286724f8eb4586fce4cb640393a32eba46

mdhorn:
@reaganlo After reverting to the sha which included the fix, I'm not seeing the GTK crashes.
Can you please on your system to see that it is stable for you as well?

mdhorn:
@reaganlo Thank you for the code review and validation help!

Issue #247, opened 2019-03-05
Enable clr-installer support in published PXE images


mdhorn:
As someone who deploys Clear Linux OS on many targets, I would like to use clr-installer instead of ister as the installation mechanism.

Currently, the PXE images published on https://download.clearlinux.org/current/ still use ister.py.

chenming1986:
A few tips to make clr-installer work with pxe boot based on current installer.img. I can help on basic verification with my iPXE setup if needed.
### iPXE setup
https://clearlinux.org/documentation/clear-linux/guides/network/ipxe-install

### kernel
use the kernel in installer.img

### kernel cmdline
copy the command line from installer.img, remove `root` and `rootfstype` parameters since we don't want to mount the real disk at this time, and add `root=/dev/ram clri.descriptor=http://192.168.1.1/clr-installer.yaml` to the cmdline. `root=/dev/ram` is neccessary, otherwise a few services and clr-installer itself will fail due to not able to write temp/configuration files

### initrd
include all the contents from the root partition of installer.img, create a symlink in the root of initrd with the name **init** and link it to **/usr/lib/systemd/systemd**, add `ExecStartPre=/usr/bin/sleep 10` to the **Service** section in **/usr/lib/systemd/system/clr-installer.service** (this is just a workaround since currently clr-installer may start before the network is ready and it will fail to fetch the installation descriptor file via network)

### iPXE configuration file
The file may look like
```
#!ipxe
kernel linux quiet console=tty0 console=ttyS0,115200n8 initrd=initrd root=/dev/ram init=/usr/bin/initra-desktop initcall_debug tsc=reliable no_timer_check noreplace-smp kvm-intel.nested=1 intel_iommu=igfx_off cryptomgr.notests rcupdate.rcu_expedited=1 rcu_nocbs=0-64 rw nomodeset i915.modeset=0 clri.descriptor=http://192.168.1.1/clr-installer.yaml
initrd initrd
boot
```

qzheng527:
Thanks Ming for the sharing, it works!
My understanding is that the released PXE image should contain the clr-installer and make it boot automatically just like it is in  the intaller image.
Who is the owner of generating the PXE image? Many internal customers really requires this install method.

mdhorn:
@qzheng527  Making updates to the PXE images to use the new clr-installer instead of ister is on the roadmap, but the core installer team right now is focused on the creation of a graphical interface for the clr-installer. We maybe not start on this for a month or two.

[...]

PR #622, opened 2019-12-09
Remove delay before PXE service

mdhorn:
- Extend delay before starting the installer service

### DIFF
--- a/etc/systemd/clr-installer-provision.service
+++ b/etc/systemd/clr-installer-provision.service
@@ -19,7 +19,6 @@ Conflicts=getty@tty1.service getty@tty0.service serial-getty@ttyS0.service
 [Service]
 Type=oneshot
 ExecStartPre=/bin/bash -l -c 'grep --quiet --no-message clri.descriptor= /proc/cmdline || (echo "Missing clr-install YAML on kernel command line" ; /bin/false)'
-ExecStartPre=-/bin/bash -l -c '/usr/bin/sleep 10'
 ExecStartPre=-/bin/bash -l -c '/usr/bin/cat /etc/issue'
 ExecStart=/bin/bash -l -c 'until /usr/bin/clr-installer ; do echo "clr-installer failed...retying"; sleep 5 ; done ; /usr/bin/reboot'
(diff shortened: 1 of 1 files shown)


etc/systemd/clr-installer-provision.service
```diff
@@ -19,7 +19,7 @@ Conflicts=getty@tty1.service getty@tty0.service serial-getty@ttyS0.service
 [Service]
 Type=oneshot
 ExecStartPre=/bin/bash -l -c 'grep --quiet --no-message clri.descriptor= /proc/cmdline || (echo "Missing clr-install YAML on kernel command line" ; /bin/false)'
-ExecStartPre=-/bin/bash -l -c '/usr/bin/sleep 10'
+ExecStartPre=-/bin/bash -l -c '/usr/bin/sleep 30'
```
jwakre: What happens when the network isn't ready after 10s? Does the until loop still work?

mdhorn: _Originally posted by @ionutnechita in https://github.com/clearlinux/clr-installer/issues/247#issuecomment-562874046_
> Hi Clear Linux Team,
> 
> If we can increase the time for clr-installer-provision.service, would help me.
> ExecStartPre=/usr/bin/sleep 10 -> ExecStartPre=/usr/bin/sleep 30
> 
> 1. My DHCP server is slower.
> 2. The interface didn't start up fast for vm using vmware esxi.

@ionutnechita  
John has a good question, I would expect that the "unit" loop to work once network comes up.

mdhorn: @ionutnechita Can you share more about the failure with the current sleep at 10?
In the loop that runs until the install is successful, we should be completely restarting the networking stack which I would assume would attempt to get a DHCP address if one was not present?

mdhorn:
Fixed in Clear Installer release [2.3.7](https://github.com/clearlinux/clr-installer/releases/tag/2.3.7)
Available in Clear Linux OS 31930 or later
https://clearlinux.org/downloads

Issue #447, opened 2019-06-20
Ensure all installation step failures show up as incomplete


reaganlo:
**Describe the bug**
When certain steps in the installation page fails, the image next to it shows up as complete (ticked). Instead, it should show up as incomplete and probably in a different color.

**Expected behavior**
When certain steps in the installation page failure, the image next to it shows up as complete (ticked). Instead it should show up as incomplete and probably in a different color.

mdhorn:
Fixed in release [2.4.0](https://github.com/clearlinux/clr-installer/releases/tag/2.4.0)

mdhorn:
Released in Clear Linux OS 32380

PR #631, opened 2020-01-08
Provide fail reasons

mdhorn:
Fixes Issue: #447

- Add which system requirement failed using system check
- Ensure all fail reason are passed back to UI
- Ensure GUI denotes the step with a fail icon


mdhorn:
@jwakre  Thanks John.

Issue #217, opened 2019-02-13
Consider to have a color blindness mode for all visual interfaces.


mbelluzzo:
Some users are finding very difficult to use clr-installer due the lack of a color blindness mode.

mdhorn:
Can you quantify this?  As it will impact the amount of effort and possible timeline.

mbelluzzo:
I can be quantified. Not sure that by me :-)

mbelluzzo:
Farther I can go to "quantify": About 8% of men and 0.5% of women suffer from color-vision deficiency.

mdhorn:
@mattcurfman This is already being tracked.

mattcurf:
Thanks @mdhorn.  I basically can't use the installer at all in its current form; have to ask people to come to my desk to see which button has focus or squint super close to screen :( recommend using a non-color cue to indicate focus, like '[' ']' brackets around item of focus.

mdhorn:
Fixed in release [2.4.0](https://github.com/clearlinux/clr-installer/releases/tag/2.4.0)

mdhorn:
Released in Clear Linux OS 32380

PR #632, opened 2020-01-08
tui: high-contrast theme

mdhorn:
Fixes Issue: #217

- Black and white theme to improve contrast

### DIFF
--- a/themes/clr-installer.theme
+++ b/themes/clr-installer.theme
@@ -1,4 +1,4 @@
-﻿// Copyright 2019 Intel Corporation
+﻿// Copyright 2020 Intel Corporation
 //
 // SPDX-License-Identifier: GPL-3.0-only
 
@@ -54,6 +54,12 @@ ListActive.EditActiveText = white
 ListActive.SelectionText = blue bold
 ListActive.SelectionBack = cyan
 
+// Highlight alt editable view
+AltEdit.EditBack = blue
+AltEdit.EditText = white
+AltEdit.EditActiveBack = cyan
+AltEdit.EditActiveText = blue
+
 // editable & listbox-like controls (interactive ones)
 EditBack       = white
 EditText       = blue bold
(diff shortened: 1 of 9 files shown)


jwakre:
When installing, I found a warning with a red background. Is this unexpected behavior for the high-contrast mode?

mdhorn:
> When installing, I found a warning with a red background. Is this unexpected behavior for the high-contrast mode?

Probably not. It looks like for the warning/error we have Red (and bold white) hardcoded in a couple of places. I'll have to refactor this. Goog catch.

mdhorn:
@jwakre The refactoring is completed and addressed the previous concerns.
All hard-coded colors are removed from the code and completely controlled by the themes.
Ready for review.

themes/clr-installer.theme
```diff
@@ -101,6 +111,8 @@ TableView=─│┼▼▲
 ManualPartition.Back = blue
 ErrorLabel.Back = blue
 ErrorLabel.Text = red bold
+WarningLabel.Back = red
+WarningLabel.Text = white
```
karthikprabhuvinod: Somehow these didn't show up in red background and white text. 
Am i missing something?
warning

jwakre: This screenshot is from the high-contrast TUI? Your comment is on the clr-installer.theme file. There is a high-contrast.theme file below.

mdhorn:
> Also, the focus in this theme, means an underline. any Specific reason we choose this? Why not over bar over the menu option?

The text-based UI doesn't allow boxes around the buttons, so we end up using light background and dark text to create a button box. In the color theme, clr-installer.theme, we can use different colors to differentiate between button boxes and the currently highlight/active text. In the high contrast/b&w, theme high-contrast.theme, we still need the light background and dark text to create the button boxes, so we instead use underline to identify the currently active text area.

themes/clr-installer.theme
mdhorn: > Somehow these didn't show up in red background and white text.
> Am i missing something?
> warning

The high contrast is using the black and white only colors in the `high-contrast.theme`.

karthikprabhuvinod:
LGTM

PR #638, opened 2020-01-21
vendor: update to latest vendor code

mdhorn:
- Switched to release 1.2.1 instead of tip of master for clui now that the required fix is in a tagger release.


karthikprabhuvinod:
@mdhorn How about the other files(vendor), are they to be checked-in?? for you it is bsd, I get a different set based on my dev machine?

PR #647, opened 2020-01-30
args: add --cbm-path

dorileo:
With the new flag we can easily instruct clr-installer to use an arbitrary
build/install, this flag eases the clr-boot-manager's CI by testing image generation
before we land new features.


mdhorn:
@dorileo thank you for making the requested change. 👍

PR #655, opened 2020-02-11
fall back to legacy bios when efi is not present

dorileo:
If the target system is not a efi system set the model to legacy bios mode.

### DIFF
--- a/clr-installer/main.go
+++ b/clr-installer/main.go
@@ -410,7 +410,7 @@ func execute(options args.Args) error {
 
 	// Run system check and exit
 	if options.SystemCheck {
-		return syscheck.RunSystemCheck(false)
+		return syscheck.RunSystemCheck(md, false)
 	}
 
 	installReboot := false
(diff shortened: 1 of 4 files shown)


karthikprabhuvinod:
I remember getting an error and getting stuck because of being on a legacy bios setting. Thanks for fixing this. I will test this soon.

dorileo:
hy @karthikprabhuvinod  I suggest you waiting this [PR](https://github.com/clearlinux/clr-boot-manager/pull/213) to be merged and land on clr before testing.

mdhorn:
@dorileo I'm wondering if it is a better idea to remove EFI check from the system check if it is really no longer a requirement -- like we did for AES.
And instead, when running interactive, we check for EFI, if not present log a warning and enable Legacy Mode.
Thoughts?

If EFI is no longer a hard requirement, we also need to update the compatibility page and checking script:
https://docs.01.org/clearlinux/latest/get-started/compatibility-check.html#compatibility-check

dorileo:
@mdhorn I believe it should work, if it doesn't work we make it work :)

mdhorn:
@dorileo working on a PR with a slightly different approach.

dorileo:
That means we cant close this one. Thank you @mdhorn

PR #659, opened 2020-02-12
kernel: Add new lts2019 kernel option

mdhorn:
- Add newest lts kernel to choose list

### DIFF
--- a/locale/es_MX/LC_MESSAGES/clr-installer.po
+++ b/locale/es_MX/LC_MESSAGES/clr-installer.po
@@ -662,6 +662,9 @@ msgstr "Kernel Linux de vanguardia adaptado a medida de metal"
 msgid "Run the latest Long Term Support (LTS) Linux Kernel"
 msgstr "Ejecute el último kernel de Linux de soporte a largo plazo (LTS)"
 
+msgid "Run the Long Term Support (LTS) 2019 Linux Kernel 5.4"
+msgstr "Ejecute el núcleo de soporte a largo plazo (LTS) 2019 Linux Kernel 5.4"
+
 msgid "Run the Long Term Support (LTS) 2018 Linux Kernel 4.19"
 msgstr "Ejecute el núcleo de soporte a largo plazo (LTS) 2018 Linux Kernel 4.19"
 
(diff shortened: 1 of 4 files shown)


mdhorn:
Waiting for lts2019 bundle to be released and default lts switched.

mdhorn:
The 2019 LTS bundle is available in the OS, so this PR is now unblocked.

_Note: The default LTS kernel is still the 2018 kernel, but that change will not be directly seen by the installer as it is only a bundle definition._

locale/es_MX/LC_MESSAGES/clr-installer.po
```diff
@@ -662,6 +662,9 @@ msgstr "Kernel Linux de vanguardia adaptado a medida de metal"
 msgid "Run the latest Long Term Support (LTS) Linux Kernel"
 msgstr "Ejecute el último kernel de Linux de soporte a largo plazo (LTS)"
 
+msgid "Run the Long Term Support (LTS) 2019 Linux Kernel 5.4"
+msgstr "Ejecute el núcleo de soporte a largo plazo (LTS) 2019 Linux Kernel 5.4"
```
karthikprabhuvinod: Should we try to parameterize this?

msgid "Run the Long Term Support (LTS) %s Linux Kernel  %s"

locale/zh_CN/LC_MESSAGES/clr-installer.po
```diff
@@ -662,6 +662,9 @@ msgstr "针对裸机定制的尖端 Linux 内核"
 msgid "Run the latest Long Term Support (LTS) Linux Kernel"
 msgstr "运行最新的长期支持 (LTS) Linux 内核"
 
+msgid "Run the Long Term Support (LTS) 2019 Linux Kernel 5.4"
+msgstr "运行长期支持 (LTS) 2019 Linux 内核 5.4"
```
karthikprabhuvinod: same here

locale/es_MX/LC_MESSAGES/clr-installer.po
mdhorn: I like the idea of parametrizing strings,  in this case, it doesn't work so well as we are loading the kernel descriptions from a JSON file and then the lookup is done based on that description:
https://github.com/clearlinux/clr-installer/blob/master/gui/pages/kernel_options.go#L93

locale/zh_CN/LC_MESSAGES/clr-installer.po
mdhorn: See above.

mdhorn:
> If I am understanding it right, the kernel is packaged as an offline bundle which gets installed by swupd during the installation. Correct?

The kernel is just another bundle selection, but in this case, the user can selected something different from the default native kernel.

mdhorn:
Fixed in clr-installer 2.4.2 first released in Clear Linux OS 32600

PR #662, opened 2020-02-13
model, log : Clean test yaml files

karthikprabhuvinod:
- backupfile needed to removed as part of model_test.go tests/valid-ister-full-physical-{timestamp}.yaml
- log/archivefile needs to be removed as part of log_test.go

### DIFF
--- a/log/log_test.go
+++ b/log/log_test.go
@@ -220,6 +220,10 @@ func TestLogTraceableError(t *testing.T) {
 }
 
 func TestFailSeek(t *testing.T) {
+	defer func() {
+		_ = os.Remove("archivefile")
+	}()
+
 	err := ArchiveLogFile("archivefile")
 	if err == nil {
 		t.Fatal("Should have failed, unseekable file")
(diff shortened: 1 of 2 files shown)


log/log_test.go
```diff
@@ -222,6 +222,9 @@ func TestLogTraceableError(t *testing.T) {
 func TestFailSeek(t *testing.T) {
 	err := ArchiveLogFile("archivefile")
 	if err == nil {
+		defer func() {
+			_ = os.Remove("./archivefile")
```
mdhorn: In theory, this is not needed as `archivefile` should not get created unless there is a bug in the code.
If we are leaving this, please drop the "./" so the filename is consistent.

log/log_test.go
```diff
@@ -231,6 +234,10 @@ func TestNoFileHandle(t *testing.T) {
 	filehandle = nil
 	err := ArchiveLogFile("archivefile")
 	if err == nil {
+		defer func() {
+			_ = os.Remove("./archivefile")
```
mdhorn: same as above

log/log_test.go
karthikprabhuvinod: Hmm ..looks like the file remains after a make check.

mdhorn: I don't think your defer should be inside the if
Perhaps before the call to ArchiveLogFile

karthikprabhuvinod: Done

log/log_test.go
karthikprabhuvinod: Done

Issue #531, opened 2019-09-18
clr-installer should support a hook for post-processing the image file


bktan8:
It would be nice for clr-installer to produce these types of images instead of having devops run a separate script to produce them.  

* cloud
* cloudguest-native
* cloudguest
* pxe
* azure
* hyperv
* vmware

mdhorn:
@bktan8 The clr-installer is used by DevOps to make those images.
The YAML files and install hooks are published for use.

I'm not sure what the ask is here?

bktan8:
clr-installer produces the raw image, but devops has to run another script to convert it to another format, say a hyperv (vhdx) image.

mdhorn:
The DevOps should be publishing the required scripts along with the other post install scripts.

mdhorn:
What we should add is a post-install hook that allows a script to be run against the final image file. That is a more generic solution to handle whatever future format come along.

mdhorn:
### Requirement
Add `post-image` hook to the YAML file which is passed the location of the final image file for any further changes that are needed.

The DevOps (or any user) can add a post-processing script to the configuration file.

mdhorn:
post-image hook is available in clr-installer 2.4.1 starting with Clear Linux OS release 32510 or higher.

PR #671, opened 2020-02-26
Post image hooks

mdhorn:
Fixes Issue: #531

- Add post-image YAML hook
- Make hooks indeterminate progress
- ensure we read the partition table when mounting loopback images


mdhorn:
@gtkramer Please take a look at this when you get a moment.

karthikprabhuvinod:
@mdhorn should we add this something to the baseline.yaml to make sure that this always run as part of make check-image and fails if someone accidentally removes the code?

mdhorn:
> Thinking of it now, with this defined, I see that we dont have a post-image defined, it assumes it to be empty. So dont really see a way of automating it in tests.
> 
I did update the baseline check target to use a post-image hook to compress the image and updated the check in the make target.

> Maybe also good to squash the commits. Upto you.
> 
In this case, I wanted to keep the three commits. 
`storage: ensure we read the partition table` is sort of related, but something new we needed to as due to recent OS update.
`controller: Make hooks indeterminate` is a separate change I've been meaning to do for a while, but it became even more noticeable when we had a single, long-running post-image hook.


> apart from a rebase, LGTM otherwise.

karthikprabhuvinod:
@mdhorn Nice LGTM :)

Issue #628, opened 2019-12-18
Allow advanced partition configuration via server yaml to use pre-existing partitions


tjad:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when my disk's partitions are created from scratch during install with server yaml configuration. ( clr-installer -c my_server.yaml )

**Describe the solution you'd like**
I would like my pre-existing partition table to be fully preserved (the installer should do absolutely no partitioning). I should be able to point to my pre-existing partitions by device label/index, and upon installation those partition structures are preserved and used for the respective CLR_ROOT, CLR_SWAP, CLR_BOOT.

**Describe alternatives you've considered**
AFAIK, no alternatives are available. I've not seen documentation regarding this. 

**Additional context**
This is for an iPXE boot. I can't use the GUI.
Ideally I use a pre-install script to partition the disk before the installer runs.

tjad:
As a suggestion, I'd change functionality such that when omitting the "size" of a specific media target child, the installer knows not to repartition. (this is implicit and may not always be clear, but does make a sensible default).

Alternatively, a "preserve_parts" option could be added to the targetMedia.

mdhorn:
@tjad Thanks for your request.

Very likely the implementation would be to use the existing [Advanced Configuration partition labels](https://docs.01.org/clearlinux/latest/get-started/bare-metal-install-desktop.html#advanced-installation) by using the CLR_* labels on partitions in the command line installer like the interactive installer.

We do have quite a backlog of feature requests to implement. To help prioritize this request, can you provide some impact information for us:
- about how many servers are you planning to deploy with this PXE feature?
- Is this a one-time use, or needed as part of a continuous deployment process?
- what is your timeline for these deployments?

Also, please keep in mind that this is an Open Source project, and we will always make time to review any submitted pull requests if you wish to pursue the implementation yourself.

mdhorn:
> @tjad Thanks for your request.
> 
> Very likely the implementation would be to use the existing [Advanced Configuration partition labels](https://docs.01.org/clearlinux/latest/get-started/bare-metal-install-desktop.html#advanced-installation) by using the CLR_* labels on partitions in the command line installer like the interactive installer.
> 
> We do have quite a backlog of feature requests to implement. To help prioritize this request, can you provide some impact information for us:
> 
> * about how many servers are you planning to deploy with this PXE feature?
> * Is this a one-time use, or needed as part of a continuous deployment process?
> * what is your timeline for these deployments?
> 
[...]

PR #673, opened 2020-02-28
storage: Enable Advanced Media Partition for CLI

mdhorn:
Fixes Issue: #628

- Allow the use of the Advanced Media Partition labels with CLI if no targetMedia is found in the YAML file.

### DIFF
--- a/massinstall/massinstall.go
+++ b/massinstall/massinstall.go
@@ -1,4 +1,4 @@
-// Copyright © 2019 Intel Corporation
+// Copyright © 2020 Intel Corporation
 //
 // SPDX-License-Identifier: GPL-3.0-only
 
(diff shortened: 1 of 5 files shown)


karthikprabhuvinod:
Quick question? Why just CLI?

mdhorn:
> Quick question? Why just CLI?

It was already been enabled in the TUI and the GUI along time ago; the request to make this available from the command line, primarily for PXE installs.

mdhorn:
Pending adding a change to remove the check for /boot if in legacy mode.

mdhorn:
@karthikprabhuvinod 
Phase one. This is being followed by a second PR that refactors the validation code.

storage/ops.go
```diff
@@ -1559,15 +1574,28 @@ func FindAdvancedInstallTargets(medias []*BlockDevice) []*BlockDevice {
 						ch.FormatPartition = true
 					}
 					clrAdded = true
+					if labelMap[labelUpper] {
+						duplicateFound = true
+						log.Error("FindAdvancedInstallTargets: Found duplicate Partition label %q", label)
```
karthikprabhuvinod: Can we move this to a function and call it like a inner function? Because we are duplicating it

massinstall/massinstall.go
```diff
@@ -125,10 +125,40 @@ func (mi *MassInstall) MustRun(args *args.Args) bool {
 func (mi *MassInstall) Run(md *model.SystemInstall, rootDir string, options args.Args) (bool, error) {
 	var instError error
 
-	// Need to ensure the partitioner knows we are running from
-	// the command line and will be using the whole disk
-	for _, curr := range md.TargetMedias {
-		md.InstallSelected[curr.Name] = storage.InstallTarget{WholeDisk: true}
+	// If there are no media defined, then we should look for
```
karthikprabhuvinod: dont strongly feel about it..but moving it a new function called is_advanced_partitioning_mode and using the return boolean to detect advanced mode or not will keep the run itself succinct. 

Upto you.

storage/ops.go
mdhorn: All of this validation related code will be removed and consolidated in a validation function in the next PR.

massinstall/massinstall.go
mdhorn: Hmm...I see what you are getting at.
But we don't really know if it is advanced mode yet, we will just check if the advanced mode labels are used only if there are no explicit media defined in the YAML file.

PR #675, opened 2020-03-04
Add complete step to the task list

mdhorn:
- Add a final step that confirms for the user that the installation has completed.


mdhorn:
Fixed in clr-installer 2.4.2 first released in Clear Linux OS 32600

PR #677, opened 2020-03-05
Check exit status of coverage testing

mdhorn:
- Added check of the exit status of the coverage testing in addition to the check for loss of coverage.
- Will require updates some tests; or removing a few root-specific tests and lowering the coverage threshold.


mdhorn:
Replaced by #680

PR #680, opened 2020-03-11
Fix tests branch

karthikprabhuvinod:
- Fix the network test to use Network Manager instead of systemd-network
- Fix the Storage test which fails sgdisk while trying to write to a partition table
- Fix the telemetry test which needed a telemetry spool directory
- Also make builds fail when one of the tests fail


karthikprabhuvinod:
@mdhorn : this is what I was talking abt:

https://github.com/clearlinux/clr-installer/blob/master/storage/storage_test.go#L552-L584

karthikprabhuvinod:
@mdhorn Looks like the sgdisk was failing as the partition creation was being skipped as those blk devices never had make partition set to true being tested for in ops.go line 453 to 455.

I have marked some part of the test while relate to advanced partitioning as obselete as i dont understand why they need to be there. Do you know?

I will wait for you to answer..

storage/storage_test.go
```diff
@@ -590,10 +590,10 @@ func TestWritePartition(t *testing.T) {
 			}
 		}()
 		bd.Name = path.Base(file)
-		part1 := &BlockDevice{Name: bd.Name + "p1", FsType: "vfat", Size: 157286400, PartitionLabel: "CLR_BOOT", Type: BlockDeviceTypePart, MountPoint: "/boot"}
-		part2 := &BlockDevice{Name: bd.Name + "p2", FsType: "swap", Size: 125829120, PartitionLabel: "CLR_SWAP", Type: BlockDeviceTypePart, MountPoint: ""}
-		part3 := &BlockDevice{Name: bd.Name + "p3", FsType: "ext4", Size: 502267904, PartitionLabel: "CLR_ROOT_F_E", Type: BlockDeviceTypePart, MountPoint: "/"}
-		part4 := &BlockDevice{Name: bd.Name + "p4", FsType: "ext4", Size: 502267904, PartitionLabel: "CLR_MNT_/home", Type: BlockDeviceTypeCrypt, MountPoint: "/home"}
+		part1 := &BlockDevice{Name: bd.Name + "p1", FsType: "vfat", Size: 157286400, PartitionLabel: "CLR_BOOT", Type: BlockDeviceTypePart, MountPoint: "/boot", MakePartition: true}
```
mdhorn: Good catch, this is why the partiton were not created.

storage/storage_test.go
```diff
@@ -590,10 +590,10 @@ func TestWritePartition(t *testing.T) {
...
+		part2 := &BlockDevice{Name: bd.Name + "p2", FsType: "swap", Size: 125829120, PartitionLabel: "CLR_SWAP", Type: BlockDeviceTypePart, MountPoint: "", MakePartition: true}
+		part3 := &BlockDevice{Name: bd.Name + "p3", FsType: "ext4", Size: 502267904, PartitionLabel: "CLR_ROOT_F_E", Type: BlockDeviceTypePart, MountPoint: "/", MakePartition: true}
```
mdhorn: We should drop the _E here as we do not currently allow encryption during advanced.
```suggestion
		part3 := &BlockDevice{Name: bd.Name + "p3", FsType: "ext4", Size: 502267904, PartitionLabel: "CLR_ROOT_F", Type: BlockDeviceTypePart, MountPoint: "/", MakePartition: true}
```

storage/storage_test.go
```diff
@@ -631,15 +631,17 @@ func TestWritePartition(t *testing.T) {
...
-		if len(found) > 0 {
-			t.Fatalf("Should not have found any advanced targets %+v", found)
-		}
-
-		if !AdvancedPartitionsRequireEncryption(bds) {
-			t.Fatalf("Advanced targets should require encryption")
-		}
+		/*
+			found := FindAdvancedInstallTargets(bds)
+			if len(found) > 0 {
```
mdhorn: We should leave these checks, but the logic now needs to be reversed.
```suggestion
			if len(found) == 0 {
```

storage/storage_test.go
```diff
@@ -631,15 +631,17 @@ func TestWritePartition(t *testing.T) {
...
+				t.Fatalf("Should not have found any advanced targets %+v", found)
```
mdhorn:
```suggestion
				t.Fatalf("Should have found any advanced targets %+v", found)
```

storage/storage_test.go
```diff
@@ -631,15 +631,17 @@ func TestWritePartition(t *testing.T) {
...
+			}
+		*/
+		/*
+			if !AdvancedPartitionsRequireEncryption(bds) {
```
mdhorn:
```suggestion
			if AdvancedPartitionsRequireEncryption(bds) {
```

storage/storage_test.go
```diff
@@ -631,15 +631,17 @@ func TestWritePartition(t *testing.T) {
...
+				t.Fatalf("Advanced targets should require encryption")
```
mdhorn:
```suggestion
				t.Fatalf("Advanced targets should not enable encryption")
```

network/network_test.go
```diff
@@ -208,7 +208,7 @@ func TestApply(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	etcDir := filepath.Join(dir, systemdNetworkdDir)
+	etcDir := filepath.Join(dir, networkManagerDir)
```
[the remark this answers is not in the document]
karthikprabhuvinod: Thanks Done

storage/storage_test.go
karthikprabhuvinod: Done

storage/storage_test.go
karthikprabhuvinod: Done

storage/storage_test.go
karthikprabhuvinod: Done

storage/storage_test.go
karthikprabhuvinod: Done

storage/storage_test.go
karthikprabhuvinod: Done

karthikprabhuvinod:
@mdhorn Addressed all concerns..Thanks for those remarks especially to keep advanced partition checkk. Please take a relook

karthikprabhuvinod:
@mdhorn I added the fix ...now those tests are passing..but a telmetry test i failing as it doesnt have a /var/pool/telemetry directory to copy records to. Any idea what needs to be done for this?

karthikprabhuvinod:
i fixed the spool by creating a directory during the test

telemetry/telemetry_test.go
```diff
@@ -361,6 +361,14 @@ func TestCopyRecords(t *testing.T) {
 		_ = os.RemoveAll(dir)
 	}()
 
+	_, err = os.Stat(telemetrySpoolDir)
```
mdhorn: If you are creating the directory, then you need to clean it up as well.
NOTE: The `TestLogRecord` test below also expects the directory to exist so will need a similar or shared fix.

The other approach I considered was skipping the test if the spool directory is missing:

```
diff --git a/telemetry/telemetry_test.go b/telemetry/telemetry_test.go
index f4da46aacbe3..65714267d7d9 100644
--- a/telemetry/telemetry_test.go
+++ b/telemetry/telemetry_test.go
@@ -347,6 +347,10 @@ func TestCopyRecords(t *testing.T) {
                t.Skip("Not running as 'root', skipping record copy test")
        }

+       if exists, err := utils.FileExists(telemetrySpoolDir); err == nil && !exists {
+               t.Skip("Not running due to missing telemetry bundle, skipping record copy test")
+       }
+
        url = "http://www.google.com"
        if err := telem.SetTelemetryServer(url, tid, policy); err != nil {
                t.Fatalf("Setting telemetry server (%q) should not return error: %s\n", url, err)
@@ -387,6 +391,10 @@ func TestLogRecord(t *testing.T) {
        if !utils.IsRoot() {
                t.Skip("Not running as 'root', skipping log record test")
        }
+       if exists, err := utils.FileExists(telemetrySpoolDir); err == nil && !exists {
+               t.Skip("Not running due to missing telemetry bundle, skipping log record test")
+       }
+
        if err := telem.LogRecord("success", 2, "Hello"); err != nil {
                t.Logf("TestLogRecord failed with %s", err)
        }
```

But you get more coverage with your solution.

karthikprabhuvinod: Okay , I think I will clean up the directory with a defer as the code gets tested

Issue #530, opened 2019-09-18
Refactor storage code to support additional features 


mdhorn:
The code for storage has become overly complex and needs to refactored to more easily enable additional features like LVM and support encryption on multiple file systems.

**Describe the solution you'd like**
The modules need to be more logically structured and remove the parallel structures like blockdevice and targetmedia.

This should be done before the other storage-related backlog items.

karthikprabhuvinod:
**Some notes:**

1.) In the storage.go I see two structs i.e: BlockDevice and BlockDeviceYaml. Can we merge them?

[https://github.com/clearlinux/clr-installer/blob/master/storage/storage.go#L37-L82](url)

2.) Looks like an good idea to keep PartitionPartition a field of Blockdevice as it natural to think of a blockdevice to keep track of its parts

mdhorn:
> https://github.com/clearlinux/clr-installer/blob/master/storage/storage.go#L37-L82

Very like that the required YAML tag could be added to the other struct.

mdhorn:
Fixed in clr-installer 2.6.1, Clear Linux OS 33300 or higher.

PR #686, opened 2020-03-24
Cleanup of storage code

karthikprabhuvinod:
1.) Move some code to test so that they are not part of executable
2.) Remove unused code

#530 

1.) Move some code to test so that they are not part of executable
2.) Remove unused code

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -12,7 +12,6 @@ import (
 	"math"
 	"path/filepath"
 	"regexp"
-	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


karthikprabhuvinod:
@mdhorn Okay. ya you are correct, The code is run from test runner and not reachable through executable's main. But still, the code should be in Test files correct? Only the code which is required for executable should be in  storage.go and ops.go.

https://github.com/clearlinux/clr-installer/pull/686#pullrequestreview-380470152

I think its okay to include test helper functions inside test files and they dont have to be _test functions.

mdhorn:
@karthikprabhuvinod Okay, here is the history; these functions were not created for driving test cases but were written for use by the TUI when the TUI had a built-in interface for disk partitioning.
For reference, you can use git blame to see where/why the code was added

```
git blame storage/storage.go
a58eb66e28d12 (Mark D Horn     2018-11-18 12:48:40 -0800  971) func IsValidLabel(label string, fstype string) string {
git show a58eb66e28d12
```

So, if these functions are truly not being used by anything in the main code, and are only exercised by test code, then we should drop the functions and the associated tests.

Issue #517, opened 2019-09-13
Installer enhancement to support clr-boot-manager updates


mdhorn:
**Is your feature request related to a problem? Please describe.**
clr-boot-manager will be making updates to better support booting non-UEFI systems.
Once these changes are released in c-b-m, the installer will need to be updated to take advantage of them.

**Describe the solution you'd like**
- Remove the hard requirement for UEFI to be enabled on the system (#666)
- Ensure the updates are made for the image creator (massinstaller), the UI, and as part of the Advanced Disk configuration

**Additional context**
- Ensure the clr-installer documents and tutorials are updated

dorileo:
c-b-m 3.2 does support the following:
+ "partitionless" /boot - we'll know how to update kernels on systems where /boot hasn't a dedicated partition;
+ /boot may be vfat or ext[2-4] formatted
+ non-EFI systems bootloaders

With the clr-installer commit eb898789bf02825b71197b23fe90c78e888181f5 we already support creating images - with massinstaller - with either vfat or ext[2-4] /boot partition.

All the remaining integrations still need to be implemented.

mdhorn:
Released in [2.4.3](https://github.com/clearlinux/clr-installer/releases/tag/2.4.3) in Clear Linux OS 32750 or higher.

PR #692, opened 2020-03-26
Refactor disk validation

mdhorn:
Fixes Issue: #517

- Enable new override functionality for media size
- Refactor and simply partition validation code
- Update partitioning support in legacy mode

### DIFF
--- a/gui/pages/disk_config.go
+++ b/gui/pages/disk_config.go
@@ -89,7 +89,7 @@ func (disk *DiskConfig) advancedButtonToggled() {
 	disk.encryptCheck.SetActive(false) // Force off for Advance as not support yet
 
 	results := storage.DesktopValidateAdvancedPartitions(disk.model.TargetMedias,
-		disk.model.LegacyBios, disk.model.SkipValidationSize)
+		disk.model.LegacyBios, disk.model.SkipValidationSize, disk.model.SkipValidationAll)
 	if len(results) > 0 {
 		disk.model.ClearInstallSelected()
 
@@ -1175,7 +1175,8 @@ func (disk *DiskConfig) GetConfiguredValue() string {
 	tm := disk.model.TargetMedias
 
 	if disk.isAdvancedSelected {
-		results := storage.DesktopValidateAdvancedPartitions(tm, disk.model.LegacyBios, disk.model.SkipValidationSize)
+		results := storage.DesktopValidateAdvancedPartitions(tm, disk.model.LegacyBios,
+			disk.model.SkipValidationSize, disk.model.SkipValidationAll)
 		if len(results) > 0 {
 			disk.model.ClearInstallSelected()
 			disk.model.TargetMedias = nil
--- a/gui/pages/disk_config.go
+++ b/gui/pages/disk_config.go
@@ -88,7 +88,8 @@ func (disk *DiskConfig) advancedButtonToggled() {
 	disk.encryptCheck.SetSensitive(storage.AdvancedPartitionsRequireEncryption(disk.model.TargetMedias))
 	disk.encryptCheck.SetActive(false) // Force off for Advance as not support yet
 
-	results := storage.DesktopValidateAdvancedPartitions(disk.model.TargetMedias)
+	results := storage.DesktopValidateAdvancedPartitions(disk.model.TargetMedias,
+		disk.model.LegacyBios, disk.model.SkipValidationSize)
 	if len(results) > 0 {
 		disk.model.ClearInstallSelected()
 
@@ -1174,7 +1175,7 @@ func (disk *DiskConfig) GetConfiguredValue() string {
 	tm := disk.model.TargetMedias
 
 	if disk.isAdvancedSelected {
-		results := storage.DesktopValidateAdvancedPartitions(tm)
+		results := storage.DesktopValidateAdvancedPartitions(tm, disk.model.LegacyBios, disk.model.SkipValidationSize)
 		if len(results) > 0 {
 			disk.model.ClearInstallSelected()
 			disk.model.TargetMedias = nil
--- a/gui/pages/disk_config.go
+++ b/gui/pages/disk_config.go
@@ -918,8 +918,12 @@ func (disk *DiskConfig) buildMediaLists() error {
 		log.Error("Failed to find storage media for install during save: %s", err)
 	}
 
-	disk.safeTargets = storage.FindSafeInstallTargets(storage.MinimumDesktopInstallSize, disk.devs)
-	disk.destructiveTargets = storage.FindAllInstallTargets(storage.MinimumDesktopInstallSize, disk.devs)
+	minSize := storage.MinimumDesktopInstallSize
+	if disk.model.SkipValidationSize {
+		minSize = 0
+	}
+	disk.safeTargets = storage.FindSafeInstallTargets(minSize, disk.devs)
+	disk.destructiveTargets = storage.FindAllInstallTargets(minSize, disk.devs)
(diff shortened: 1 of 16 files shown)


storage/storage.go
```diff
@@ -428,18 +428,6 @@ func (bd *BlockDevice) IsAvailable() bool {
...
 }
 
-// ConfigStatus is the status type for Configuration Status of the installation media
-type ConfigStatus int
-
-const (
-	// ConfiguredNone indicates no installation media has been configured
-	ConfiguredNone ConfigStatus = iota
-	// ConfiguredPartial indicates only some of the required partition are configured
-	ConfiguredPartial
```
karthikprabhuvinod: 👍 .Thanks

tui/tui.go
```diff
@@ -103,6 +103,12 @@ func (tui *Tui) Run(md *model.SystemInstall, rootDir string, options args.Args)
 		tui.model.CopySwupd = true
 	}
 
+	// Check if we can boot EFI
+	if !utils.HostHasEFI() {
```
karthikprabhuvinod: Why do we do this check only for tui?

gui/pages/disk_config.go
```diff
@@ -88,7 +88,8 @@ func (disk *DiskConfig) advancedButtonToggled() {
 	disk.encryptCheck.SetSensitive(storage.AdvancedPartitionsRequireEncryption(disk.model.TargetMedias))
 	disk.encryptCheck.SetActive(false) // Force off for Advance as not support yet
 
-	results := storage.DesktopValidateAdvancedPartitions(disk.model.TargetMedias)
+	results := storage.DesktopValidateAdvancedPartitions(disk.model.TargetMedias,
+		disk.model.LegacyBios, disk.model.SkipValidationSize)
```
karthikprabhuvinod: Looks like DiskConfig type has access to a model struct from which we can extract these bits of info. Maybe just pass the DiskConfig  and we can extract all these values in DesktopValidateAdvancedPartitions itself

massinstall/massinstall.go
```diff
@@ -154,6 +154,19 @@ func (mi *MassInstall) Run(md *model.SystemInstall, rootDir string, options args
 		}
 		if isAdvancedSelected {
 			log.Debug("Mass installer operating in Advanced Disk Partition Mode.")
+			var results []string
+			if md.IsDesktopInstall() {
```
karthikprabhuvinod: I am not sure why the check for IsDesktopInstall is required in massinstall.go? Am i missing something?

storage/ops.go
```diff
@@ -1793,52 +1796,42 @@ func validatePartitions(rootSize uint64, medias []*BlockDevice, legacyBios bool,
...
-					results = append(results, warning)
-					log.Warning("validatePartitions: legacyMode, invalid fsType: %s %s %+v",
-						rootBlockDevice.FsType, warning, rootBlockDevice)
-				}
-				if rootBlockDevice.Type == BlockDeviceTypeCrypt {
-					warning := utils.Locale.Get("Encryption of %s is not supported", rootLabel)
-					results = append(results, warning)
-					log.Warning("validatePartitions: legacyMode without /boot, %s %+v",
-						warning, rootBlockDevice)
+		if skipAll {
```
karthikprabhuvinod: why did we break it down one level.... it could have remained the same as before:   if legacyBios && SkipAll

massinstall/massinstall.go
```diff
@@ -130,27 +130,41 @@ func (mi *MassInstall) Run(md *model.SystemInstall, rootDir string, options args
...
 		// Need to ensure the partitioner knows we are running from
 		// the command line and will be using the whole disk
 		for _, curr := range md.TargetMedias {
 			md.InstallSelected[curr.Name] = storage.InstallTarget{WholeDisk: true}
 			log.Debug("Mass installer using defined media in YAML")
 		}
 
-		if md.SkipValidationAll {
-			log.Warning("massinstall: Skipping all partition checks due to SkipValidationAll")
+		if md.IsDesktopInstall() {
```
karthikprabhuvinod: Same here...Why perform IsDesktopInstall check?

tui/tui.go
mdhorn: We check in both the TUI and the GUI https://github.com/clearlinux/clr-installer/blob/master/gui/window.go#L195
The problem being fixed here is that in the TUI the EFI check was too late in the code; it is now moved before the partition information is processed.

massinstall/massinstall.go
mdhorn: IsDesktopInstall looks at the list of bundles to be installed. If there are desktop related bundles, we need much more disk space.
This is why there is a check for Desktop or Server, as that impacts the minimum required disk space needed for the installation regardless if we are installing from the CLI, TUI, or GUI.

massinstall/massinstall.go
mdhorn: IsDesktopInstall looks at the list of bundles to be installed. If there are desktop related bundles, we need much more disk space.
This is why there is a check for Desktop or Server, as that impacts the minimum required disk space needed for the installation regardless if we are installing from the CLI, TUI, or GUI.

storage/ops.go
mdhorn: No, that logic doesn't work (that was my first approach which didn't cover all the cases)
We need:

- No Boot Found, Skip Validation Requested, **and** Legacy BIOS mode (We still need to check root partition)
- No Boot Found, Skip Validation Requested, **and** NOT Legacy BIOS mode (Error: Must have /boot if not Legacy mode)
- No Boot Found, NOT Skip Validation Requested (Error: Must have /boot if not Legacy mode)

gui/pages/disk_config.go
mdhorn: The storage.DesktopValidateAdvancedPartitions is used by CLI, TUI, and the GUI, so we can't pass a structure unique to just one of those.
We can't just pass a pointer to the model or we get recursive inclusions.

mdhorn:
@karthikprabhuvinod Please see my responses.

gui/pages/disk_config.go
karthikprabhuvinod: Cant we only pass pointer to the model(only this parameter) and unpack things from it if we need it later?

massinstall/massinstall.go
karthikprabhuvinod: @mdhorn Thanks for explaining, maybe we need to change the name of the method itself
ContainsDesktopBundles() ?

Leave it upto you

storage/ops.go
karthikprabhuvinod: Is it possible to move sections of code to another helper functions, This function is getting increasily tough to read.

massinstall/massinstall.go
karthikprabhuvinod: Same as above

gui/pages/disk_config.go
mdhorn: > Cant we only pass pointer to the model(only this parameter) and unpack things from it if we need it later?

We can't just pass a pointer to the model or we get recursive inclusions; as stated above.

massinstall/massinstall.go
mdhorn: I'll rename it to IsTargetDesktopInstall.

storage/ops.go
mdhorn: @karthikprabhuvinod  Hence #530

karthikprabhuvinod: @mdhorn Sure will take care of it during #530

PR #693, opened 2020-03-26
kernel: Drop extra LTS kernels

mdhorn:
- Only offer the latest LTS kernel as an alternative

### DIFF
--- a/etc/kernels.json
+++ b/etc/kernels.json
@@ -10,21 +10,6 @@
       "name": "LTS Latest",
       "desc": "Run the latest Long Term Support (LTS) Linux Kernel",
       "note": "THIS BUNDLE is auto-selected for VirtualBox in model.go"
-    },
-		{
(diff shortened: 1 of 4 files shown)


bryteise:
LGTM

karthikprabhuvinod:
Changewise looks good to me but any reason why are we doing this? Why not allow more lts versions? I am guessing this is only during installation but the user can choose to change the kernel if he likes after installlation correct?

mdhorn:
> Changewise looks good to me but any reason why are we doing this? Why not allow more lts versions? I am guessing this is only during installation but the user can choose to change the kernel if he likes after installlation correct?

Yes, they can change after the installation.
See https://github.com/clearlinux/distribution/issues/1857 for context.

mdhorn:
Released in [2.4.3](https://github.com/clearlinux/clr-installer/releases/tag/2.4.3) in Clear Linux OS 32750 or higher.

PR #694, opened 2020-03-28
tests: Add new test to build all DevOps Images

mdhorn:
- New validation/QA code to attempt to build all images published by our DevOps

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ build-tui:
 		-ldflags="-X github.com/clearlinux/clr-installer/model.Version=${VERSION} \
 		-X github.com/clearlinux/clr-installer/model.BuildDate=${BUILDDATE}" \
 		${GO_PACKAGE_PREFIX}/clr-installer
-	mv ${GOPATH}/bin/clr-installer ${GOPATH}/bin/clr-installer-tui
+	mv -f ${GOPATH}/bin/clr-installer ${GOPATH}/bin/clr-installer-tui
 
 build-gui:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -203,9 +203,44 @@ check-image: build-tui
...
+check-all-images-screen: check-all-images-core
+	@echo "Lauching screen sessions..."; \
+	echo "Use <control>-a<controld>-d to detach and leave running."; \
+	echo "	sudo screen -R # to recover session"
+	@sleep 2
+	@screen_log="screen-check-all-images.log"; \
+	if [ -f $${screen_log} ] ; then \
+		mv -f $${screen_log} $${screen_log}.save; \
+	fi; \
+	cd ${top_srcdir} ; screen -L -Logfile $${screen_log} tests/check-all-images.sh . ; \
```
karthikprabhuvinod: Maybe the change directory into the top_srcdir is redundant as a user needs already needs to exist there to invoke make target

mdhorn:
> @mdhorn Hey Mark, I ran the shell script through shell check and it gave me a bunch of errrors. Please could you take a look at those before merging?
> 

Two notes, 1. the output here is from a different script, 2. the checkers in question would be nice to have, but is not currently available in Clear Linux OS.

I have address the items flagged by the checker, but these scripts are not a part of the distribution, and until the check is a part of the OS, we shall not require it's usage to gate PRs.

mdhorn:
Released in [2.4.3](https://github.com/clearlinux/clr-installer/releases/tag/2.4.3) in Clear Linux OS 32750 or higher.

PR #696, opened 2020-04-01
lint: Only ignore the few problem functions

mdhorn:
- Instead of skipping whole files, just exclude the few complex functions.


karthikprabhuvinod:
I like this better

PR #700, opened 2020-04-08
Lint: Add a target to check source line length

karthikprabhuvinod:

- Add a target lint-lll to check sourc ecode line length
- Make changes to code whereebver it fails


mdhorn:
Thank you for enabling this linter. Do we really want to force things down to 80? Maybe 100?
We should have a discussion with the mixer team. Do we only care about go files only?
Maybe there are good reasons for an occasional long line?
`-l 100 -g -e "TODO|FIXME" 
`

karthikprabhuvinod:
@mdhorn : So I think the default is 120. Thats we are using. Since you use vim, this will help you a lot. Also makes reviewing and reeading code easier on github.

> Do we only care about go files only?
Good question. I would opt choosing all code to follow it. Somehow using a go linter for all source code files 

> Maybe there are good reasons for an occasional long line?
There were places I didnt want to do the line shortening like in storage_test where we have json output which looks exactly like a replica from 
`lsblk -J -O /dev/sdx`.

mdhorn:
> @mdhorn : So I think the default is 120. Thats we are using. Since you use vim, this will help you a lot. Also makes reviewing and reeading code easier on github.
> 
> > Do we only care about go files only?
> > Good question. I would opt choosing all code to follow it. Somehow using a go linter for all source code files
> 
> > Maybe there are good reasons for an occasional long line?
> > There were places I didnt want to do the line shortening like in storage_test where we have json output which looks exactly like a replica from
> > `lsblk -J -O /dev/sdx`.

The Git page says the default is 80 characters, but I didn't test this.
I sort of like the suggestion of 100 characters and excluding lines with TODO or FIXME
`-l 100  -e "TODO|FIXME"`
But we should discuss this in IRC with the mixer team so we can agree on a team standard.

karthikprabhuvinod:
> 
> The Git page says the default is 80 characters, but I didn't test this.
> I sort of like the suggestion of 100 characters and excluding lines with TODO or FIXME
> `-l 100 -e "TODO|FIXME"`
> But we should discuss this in IRC with the mixer team so we can agree on a team standard.


@mdhorn  I read the golangci documentation, to override the defaults, we will need to create golangci.yaml or json file to over-ride options for individual linters!

karthikprabhuvinod:
@mdhorn @reaganlo 
1.) I searched and thought hard, but i couldn't find a way to exclude on basis of exclude strings we define. Looks like --exclude is used to exclude error messages stemming from error messages and not user defined tags. Instead we can use something like

//nolint: lll // WONTFIX
//nolint: lll // TODO
//nolint: lll // FIXME

**It helps linter skip the smallest possible construct(method, var etc) and skip a specific lint test on it**

2.) Also for tab-width check for lll works,  but  it creates lots of things to fix. I can still go and fix this but these will be all manual because lll has to auto-fix :(

@mdhorn: let me know

karthikprabhuvinod:
@mdhorn I added tab-width fix..looks like I fixes werent a whole lot.

.golangci.yml
```diff
@@ -0,0 +1,23 @@
+run:
+  timeout: 10m
+
+linters-settings:
+  lll:
+    tab-width: 4
```
mdhorn, line 6: Please add the `    line-length: 120` as well so we can see what our default should be. Also, this will avoid problems if the default changes at a future date.

.golangci.yml
```diff
@@ -0,0 +1,23 @@
...
+    - gocyclo
+    - gofmt
+    - golint
+    - deadcode
+    - varcheck
+    - structcheck
+    - unused
+    - vetshadow
+    - errcheck
+    - lll
```
mdhorn, line 23: Please add a newline

gui/pages/kernel_options.go
```diff
@@ -7,12 +7,13 @@ package pages
 import (
 	"github.com/clearlinux/clr-installer/gui/common"
 
+	"strings"
```
mdhorn, line 10: Please make "strings" first.
The imports should always be

1.  Built-in Language
2. Vendor Code
3. This application's packages

model/model_ister.go
```diff
@@ -123,12 +123,14 @@ func JSONtoYAMLConfig(cf string) (*SystemInstall, error) {
 			case "virtual":
 				bd.Name = curr.Disk
 				bd.Type = storage.BlockDeviceTypeLoop
-				sa.Name = strings.TrimSuffix(curr.Disk, filepath.Ext(curr.Disk)) // remove any extensions from alias name
+				sa.Name = strings.TrimSuffix(curr.Disk,
+					filepath.Ext(curr.Disk)) // remove any extensions from alias name
```
mdhorn, line 127: If and end of line comment is making a line too long, I prefer the solution used above -- move the comment before the line of code.

model/model_ister.go
```diff
@@ -123,12 +123,14 @@ func JSONtoYAMLConfig(cf string) (*SystemInstall, error) {
...
 				sa.File = curr.Disk
 			case "physical":
 				bd.Name = curr.Disk
 				bd.Type = storage.BlockDeviceTypeDisk
-				sa.Name = strings.TrimSuffix(curr.Disk, filepath.Ext(curr.Disk)) // remove any extensions from alias name
+				sa.Name = strings.TrimSuffix(curr.Disk,
+					filepath.Ext(curr.Disk)) // remove any extensions from alias name
```
mdhorn, line 133: If and end of line comment is making a line too long, I prefer the solution used above -- move the comment before the line of code.

swupd/swupd_test.go
```diff
@@ -169,6 +169,7 @@ func TestProcess(t *testing.T) {
 	}
 
 	// "progress" messages should be processed correctly
+	//nolint: lll // WONTFIX
```
mdhorn, line 172: Does this work on a single line?
I thought we found that the nolint only applied to a variable, function, or the module?

swupd/swupd_test.go
```diff
@@ -180,10 +181,13 @@ func TestProcess(t *testing.T) {
 	if mp.output != "" {
 		t.Fatal("Message processed incorrectly. Expected: '', Actual:", mp.output)
 	}
+	//nolint: lll // WONTFIX
```
mdhorn, line 184: What is the scope of this //nolint?

gui/pages/kernel_options.go
karthikprabhuvinod: Done

model/model_ister.go
karthikprabhuvinod: Done

model/model_ister.go
karthikprabhuvinod: Done

.golangci.yml
karthikprabhuvinod: Done

.golangci.yml
karthikprabhuvinod: Done

swupd/swupd_test.go
karthikprabhuvinod: scope is limited to variable following //nolint: jsonMSG  and only for lll

swupd/swupd_test.go
karthikprabhuvinod: Yes it does... it is being applied to the variable jsonMsg

mdhorn:
Fixed in installer release [2.5.0](https://github.com/clearlinux/clr-installer/releases/tag/2.5.0) found in Clear Linux OS release 32980 or greater.

PR #709, opened 2020-04-15
Support software raid in Advanced partitioning flow

karthikprabhuvinod:

 Partially Addresses : #8 

- In Advanced partitioning flow, allow the use of raid devices as targets for installation. Make sure that boot partition needs to be on non-raid device and non-lvm device

### DIFF
--- a/storage/block_devices_ops.go
+++ b/storage/block_devices_ops.go
@@ -544,7 +544,9 @@ func partitionUsingParted(bd *BlockDevice, dryRun *[]string, wholeDisk bool) err
 
 // WritePartitionTable writes the defined partitions to the actual block device
 func (bd *BlockDevice) WritePartitionTable(legacyBios bool, wholeDisk bool, dryRun *[]string) error {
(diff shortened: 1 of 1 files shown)


karthikprabhuvinod:
@mdhorn : Not asking for review for merge, Just to get your thoughts for this?

mdhorn:
> @mdhorn : Not asking for review for merge, Just to get your thoughts on this?

This is a good start, and I used this PR to create an install -- which of course doesn't boot, but gives me a testing framework.
The changes you should make are moving those checks from WritePartition table to the Findxxxx routines. The pre-existing check was mainly a sanity check for writing the partitions. The checking should be done in the Findxxx routines to screen out.
We should be flagging that /boot must be a physical disk there so the message is conveyed back to the user.

mdhorn:
Fixed in installer release [2.5.0](https://github.com/clearlinux/clr-installer/releases/tag/2.5.0) found in Clear Linux OS release 32980 or greater.

PR #718, opened 2020-04-25
Fix setting the boot partition

mdhorn:
- Fixes a problem where the boot partition may not be set correctly if boot and root are on different media.
- Add missing translation

### DIFF
--- a/storage/block_devices_ops.go
+++ b/storage/block_devices_ops.go
@@ -328,51 +328,19 @@ func (bd *BlockDevice) writePartitionLabel(wholeDisk bool) error {
 	return nil
 }
 
-// handlerPartitionLabelWrite is a helper function to WritePartitionTable
-// prepares performs a few checks before updating the label for the partition
-func (bd *BlockDevice) handlerPartitionLabelWrite(legacyBios bool, wholeDisk bool, dryRun *[]string) error {
-	//write the partition label
-	if dryRun != nil {
-		if legacyBios {
-			*dryRun = append(*dryRun, bd.Name+": "+utils.Locale.Get(LegacyModeWarning))
-		}
-
-		// Check if there is a boot partition
-		bootFound := false
-		for _, curr := range bd.Children {
-			// First, check if we have the standard /boot partition
-			if curr.MountPoint == "/boot" {
-				bootFound = true
-			}
-		}
-
-		if !bootFound && legacyBios {
-			*dryRun = append(*dryRun, bd.Name+": "+utils.Locale.Get(LegacyNoBootWarning))
-		}
-
-		if wholeDisk {
-			*dryRun = append(*dryRun, bd.Name+": "+utils.Locale.Get(PartitioningWarning))
-		}
-	} else {
-		if err := bd.writePartitionLabel(wholeDisk); err != nil {
-			return err
-		}
-	}
-
-	return nil
-}
-
-// adjustFileSystems is a helper function to WritePartitionTable takes a prepared
+// setPartitionGUIDs is a helper function to WritePartitionTable takes a prepared
 // guid map of GUIDS->device names and uses sgdisk to update the
 // guid partition table for the disk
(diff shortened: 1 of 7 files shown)


karthikprabhuvinod:
@mdhorn sorry about the delay...will review first thing in morning

storage/block_devices_ops.go
```diff
@@ -2076,6 +2063,116 @@ func GetAdvancedPartitions(medias []*BlockDevice) []string {
...
+	// Clear Linux OS only supports booting from a top level
+	// block device; RAID, LVM, encryption, etc are not supported
+	for _, bd := range medias {
+		for _, curr := range bd.Children {
+			// We have the standard /boot partition
+			if curr.MountPoint == "/boot" {
+				if bootBlockDevice != nil {
+					return errors.Errorf(logFormatError("Found multiple %s partition names", curr.MountPoint))
+				}
+				bootParent = bd
```
karthikprabhuvinod, line 2100: I would suggest that we use something like the https://github.com/clearlinux/clr-installer/blob/master/storage/block_devices_filters.go#L20 function to find blockdevices that meet a certain condition....I am bringing in a few more that help in my make destructive patch.

I think your condition wont be satisfied with it alone. I have seen at many places where we want parent and a child returned if certain condition is met. It Simplifies the code and prevents repetition

storage/storage_test.go
```diff
@@ -939,6 +939,11 @@ var lsblkOutput string = `{
             {"name": "sdg3", "path": "/dev/sdg3", "size": "8G",    "type": "part", "fstype": "ext4", "partlabel": "CLR_ROOT", "mountpoint": "/"},
             {"name": "sdg4", "path": "/dev/sdg4", "size": "8G",    "type": "part", "fstype": "ext4", "partlabel": "CLR_MNT_/home", "mountpoint": "/home"}
          ]
+      },
+	  {"name": "sdh", "path": "/dev/sdh", "size": "2.0T", "type": "disk", "mountpoint": null,
```
karthikprabhuvinod, line 943: Not sure why this test? Is this to show we have a root partition and no explicit boot partition(simulating a legacy mode) and test this passes?

storage/block_devices_ops.go
mdhorn: Currently this is the only use case for needing both the child and parent. Find the boot partition is a special case.

mdhorn:
@karthikprabhuvinod Found the other issue you mentioned with using installer.yaml and setting the alias.

karthikprabhuvinod:
@mdhorn cool did you forcepush towards another branch?
 Looks like you changes from completions PR..check force-pushed

mdhorn:
> completions

Nope, I rebased against the upstream master as well as adding fixes.

karthikprabhuvinod:
I perfomed various tests and they seem to work well. 

**Installed with :**

```
# clear-linux-config

# installer.yaml
#
# This YAML file generates the basic TUI installer image for
# Clear Linux OS

# Switch between aliases if you want to install to an actual block device.
# i.e /dev/sda
block-devices: [
   {name: "installerboot", file: "installer.img"},
   {name: "installerother", file: "installer.img"}
]

targetMedia:
- name: ${installerboot}
  type: disk
  children:
  - name: ${installerboot}1
    fstype: vfat
    mountpoint: /boot
    size: "150M"
    type: part
- name: ${installerother}
  type: disk
  children:
  - name: ${installerother}1
    fstype: swap
    size: "256M"
    type: part
  - name: ${installerother}2
    fstype: ext4
    mountpoint: /
    size: "2.6G"
    type: part
```


**result:**

```
NAME      MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
loop0       7:0    0    15G  0 loop
├─loop0p1 259:3    0   255M  0 part
└─loop0p2 259:4    0   2.6G  0 part
loop1       7:1    0    15G  0 loop
└─loop1p1 259:1    0   149M  0 part
```


Worked well for me :)

Issue #711, opened 2020-04-15
Bash and/or zsh auto-completion script for command line installer


karthikprabhuvinod:
**Is your feature request related to a problem? Please describe.**
Currently there is no tab-completion or auto-complete feature in bash or zsh for clr-installer

**Describe the solution you'd like**
It will nice if we auto-complete for options in bash and/or zsh. Makes the CLi installer more user-friendly. 

**Describe alternatives you've considered**
Currently the best way around this to print clr-installer --help or clr-installer docs

mdhorn:
Most of our user use the GUI, then a small percentage use the TUI.
Command-line completion is really only useful for the developers, hence it was never prioritized. 😉

lebensterben:
Hey I've written one last year. But I forget to publish it....

lebensterben:
@karthikprabhuvinod 
It's kind of completed **as of last year**. But I haven't thoroughly test it.
https://github.com/lebensterben/clr-installer/tree/completions/completions

lebensterben:
Actually I invented some tricks using invisible unicode symbols to display block device for bash. (Though it's only for display you can not actually use tab-completion).

karthikprabhuvinod:
@lebensterben pretty cool.. Do you not wish to upstream this?

lebensterben:
> @lebensterben pretty cool.. Do you not wish to upstream this?

I need to update this first and test it.

lebensterben:
@karthikprabhuvinod 
Just to give you a quick update on my progress.
I was updating my shell completions for `swupd-client` and `clr-boot-manager` in the last week. Well it took longer than I thought due to a major upgrade in `swupd-client`.
I will focus on the `clr-installer` in the next few days. `zsh` part is done, and I've tested it before. I just need to update it. For `bash`, I'm still writing it. I will not use any `tricks` I mentioned before, because I found that it might be hard for other maintainers to understand what I'm doing. So I will just stick to conventions.

karthikprabhuvinod:
@lebensterben thanks for updating us 😊. That's fine. Will be happy to review your code whenever you can submit.

mdhorn:
Fixed in clr-installer 2.6.1, Clear Linux OS 33300 or higher.

PR #719, opened 2020-04-25
completions: Added Completions for bash and zsh

lebensterben:
- Added `zsh` completion script at `completions/zsh/_clr-installer`. It's
  suggested to install it at
  `/usr/share/zsh/site-functions/_clr-installer`.
- Added `bash` completion script at `completions/bash/clr-installer`. It's
  suggested to install it at
  `/usr/share/bash-completion/completions/clr-installer`

Fixes Issue: #711 

---

## Notes on Bash autocompletion script

I invented some tricks with unicode non-printable characters to provide additional functionality not supported by bash builtin completion system:

- line 32-42, will show a help message about the format of the argument to
  `--block-device` option, as 
   >Format:"alias1:filename1[,alias2:filename2,...]" BlockDevices:${blockdevice:0:-1}"
  where `$blockdevice` are generated dynamically. This message won't be
  completed.
- line 62-64, will remind the user to use `https://` urls. This message won't be
  completed as well.
- line 68-91, will first show a help message on the definition of log-levels,
  >error-->1 warning-->2 info-->3 debug--->4
  When the user type any of `e`, `w`, `i`, or `d`, and then press `<TAB>` it
  would be replaced by the integer value accordingly. When the user type any
  other character, the message persists.

The drawback of this trick is, those unicode characters would be rendered as a solid rectangle in some terminals, such as tty.

If this is not acceptable, just remove the aforementioned lines.

### DIFF
--- a/completions/bash/clr-installer
+++ b/completions/bash/clr-installer
@@ -24,7 +24,7 @@ _clr_installer() {
   _init_completion -s || return
 
   case "$prev" in
-    --archive|--copy-network|--iso|--keep-image|--reboot|--stub-image|--swupd-skip-diskspace-check|--telemetry)
+    --archive|--copy-network|--copy-swupd|--keep-image|--reboot|--swupd-skip-diskspace-check)
       COMPREPLY=($(compgen -W "true false" -- "$cur"))
       return
       ;;
@@ -43,22 +43,26 @@ _clr_installer() {
       return
       ;;
     --config)
-      COMPREPLY=($(compgen -f -X "!*.yaml" -- "$cur"))
+      _filedir yaml
       return
       ;;
     --json-yaml)
-      COMPREPLY=($(compgen -f -X "!*.json" -- "$cur"))
+      _filedir json
       return
       ;;
     --crypt-file|--log-file)
       COMPREPLY=($(compgen -f -- "$cur"))
       return
       ;;
-    --swupd-cert|--swupd-state)
+    --swupd-cert)
+      _filedir pem
+      return
+      ;;
+    --swupd-state)
       COMPREPLY=($(compgen -d -- "$cur"))
       return
       ;;
-    --swupd-contenturl|--swupd-mirror|--swupd-versionurl|--telemetry-url)
+    --swupd-url|--swupd-contenturl|--swupd-mirror|--swupd-versionurl|--telemetry-url)
       # a zero-width space before `h`, and at the last character
       opts='​https://...  '
       COMPREPLY=($(compgen -W "$opts" -- "$cur"))
@@ -82,6 +86,8 @@ _clr_installer() {
         ​d*)
           opts="4"
           COMPREPLY=($(compgen -W "$opts" -- "$opts"))
+          ;;
+        -B|--bundles)
           ;;
         *)
           # a zero-width space before each alternative
--- a/completions/bash/clr-installer
+++ b/completions/bash/clr-installer
@@ -1,0 +1,108 @@
+#!/bin/bash
+# shellcheck disable=SC2207
+# -----------------------------------------------------------------------
+#   Clear Linux OS* Installer - autocompletion script
+#
+#   Author: Lucius Hu - http://github.com/lebensterben
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, version 2 or later of the License.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# -----------------------------------------------------------------------
+
+_clr_installer() {
+  local IFS opts prev cur
+
+  _init_completion -s || return
+
+  case "$prev" in
+    --archive|--copy-network|--iso|--keep-image|--reboot|--stub-image|--swupd-skip-diskspace-check|--telemetry)
+      COMPREPLY=($(compgen -W "true false" -- "$cur"))
+      return
+      ;;
+    --block-device)
+      local blockdevice; blockdevice="$(lsblk --noheadings --raw --paths | cut -d' ' -f1 | tr '\n' ',')"
+      # The trick is to insert any UNICODE space(s), e.g. non-breaking space, before the message
+      #    so that 'word' is not likely to be matched by a user's input
+      # And add at least one other 'word' start with any UNICODE space(s)
+      #    otherwise completion system will push the only candidate to the command line
+      # The drawback is, this special character may not shown correctly, for example in `tty`, it's
+      #    it's shown as a rectangle
+
+      # here, before F and B, there is a UNICODE tag space U+E0020
+      opts="󠀠Format:\"alias1:filename1[,alias2:filename2,...]\" 󠀠BlockDevices:${blockdevice:0:-1}"
+      COMPREPLY=($(compgen -W "$opts" -- "$cur"))
+      return
+      ;;
+    --config)
+      COMPREPLY=($(compgen -f -X "!*.yaml" -- "$cur"))
+      return
+      ;;
+    --json-yaml)
+      COMPREPLY=($(compgen -f -X "!*.json" -- "$cur"))
+      return
+      ;;
+    --crypt-file|--log-file)
+      COMPREPLY=($(compgen -f -- "$cur"))
+      return
+      ;;
+    --swupd-cert|--swupd-state)
+      COMPREPLY=($(compgen -d -- "$cur"))
+      return
+      ;;
+    --swupd-contenturl|--swupd-mirror|--swupd-versionurl|--telemetry-url)
+      # a zero-width space before `h`, and at the last character
+      opts='​https://...  '
+      COMPREPLY=($(compgen -W "$opts" -- "$cur"))
+      return
+      ;;
+    --log-level)
+      # shellcheck disable=1018
+      case "$2" in
+        ​e*)
+          opts="1"
+          COMPREPLY=($(compgen -W "$opts" -- "$opts"))
+          ;;
+        ​w*)
+          opts="2"
+          COMPREPLY=($(compgen -W "$opts" -- "$opts"))
+          ;;
+        ​i*)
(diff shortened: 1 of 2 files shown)


completions/bash/clr-installer
```diff
@@ -0,0 +1,108 @@
...
+      ;;
+    --crypt-file|--log-file)
+      COMPREPLY=($(compgen -f -- "$cur"))
+      return
+      ;;
+    --swupd-cert|--swupd-state)
+      COMPREPLY=($(compgen -d -- "$cur"))
+      return
+      ;;
+    --swupd-contenturl|--swupd-mirror|--swupd-versionurl|--telemetry-url)
```
mdhorn, line 61: For consistency, we should include `--swupd-url` as well.

completions/bash/clr-installer
```diff
@@ -0,0 +1,108 @@
...
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# -----------------------------------------------------------------------
+
+_clr_installer() {
+  local IFS opts prev cur
+
+  _init_completion -s || return
+
+  case "$prev" in
+    --archive|--copy-network|--iso|--keep-image|--reboot|--stub-image|--swupd-skip-diskspace-check|--telemetry)
```
mdhorn, line 27: For consistency, we should include:
```suggestion
    --allow-insecure-http|--archive|--cfPurge|--copy-network|--copy-swupd|--high-contrast|--iso|--keep-image|--offline|--reboot|--stub-image|--swupd-clean|--swupd-skip-diskspace-check|--swupd-skip-optional|--telemetry|--tui)
```

Is there some standard to show what the default is for these true/false items? Most of these default to false, but a few default to true.

completions/bash/clr-installer
```diff
@@ -0,0 +1,108 @@
...
+      # And add at least one other 'word' start with any UNICODE space(s)
+      #    otherwise completion system will push the only candidate to the command line
+      # The drawback is, this special character may not shown correctly, for example in `tty`, it's
+      #    it's shown as a rectangle
+
+      # here, before F and B, there is a UNICODE tag space U+E0020
+      opts="󠀠Format:\"alias1:filename1[,alias2:filename2,...]\" 󠀠BlockDevices:${blockdevice:0:-1}"
+      COMPREPLY=($(compgen -W "$opts" -- "$cur"))
+      return
+      ;;
```
mdhorn, line 44: I wonder if it is reasonable or worthwhile to add --bundle with a completion similar to what swupd as?  See /usr/share/defaults/etc/profile.d/50-swupd.bash
The list would probably get too long?

completions/bash/clr-installer
```diff
@@ -0,0 +1,108 @@
...
+    --config)
```
mdhorn, line 45: Is there any way to get this to expand with a path?
I would generally use --config <path>/[tab]

Same for other file expanders

completions/bash/clr-installer
```diff
@@ -0,0 +1,108 @@
...
+      ;;
+    --json-yaml)
+      COMPREPLY=($(compgen -f -X "!*.json" -- "$cur"))
+      return
+      ;;
+    --crypt-file|--log-file)
+      COMPREPLY=($(compgen -f -- "$cur"))
+      return
+      ;;
+    --swupd-cert|--swupd-state)
```
mdhorn, line 57: `--swupd-cert` is actually a full path to a PEM file; this should probably move up with `--crypt-file` and `--log-file`

completions/bash/clr-installer
lebensterben: The list is getting from MoM file.
Does installer live ISO has the MoM file as well? (And is it in same location?)

completions/bash/clr-installer
lebensterben:
```bash
--config)
	COMPREPLY=($(compgen -f -X "!*.yaml" -- "$cur"))
	return
;;
```

I've replaced it with `_filedir yaml`, that's should be better.

completions/bash/clr-installer
lebensterben: I've replaced it with `_filedir pem`, that accepts `*.pem` file.

The `--crypt-file` and `--log-file` just accepts any file. Is there any requirement on the extension?

completions/bash/clr-installer
lebensterben: The problem is I compiled the latest `clr-installer` but it was not in my PATH. So I actually missed some of those options.

And I should actually remove some of them.
So it would be
```bash
--archive|--copy-network|--copy-swupd|--keep-image|--reboot|--swupd-skip-diskspace-check)
```

If something defaults to true, you have to specify `--foo false` to turn it off.

But in the case of all other options that accepts a boolean, they defaults to false, and you just need `--foo` to turn it on.

P.S.
Also, it's rather verbose to have `--foo true|false`.
I suggests replace the 6 options that defaults to true by
- `--archive` -> `--no-archive`
- `--copy-network` -> `--no-copy-network`
- `--copy-swupd` -> `--no-copy-swupd`
- `--keep-image` -> `--no-keep-image`
- `--reboot` -> `--no-reboot`
- `--swupd-skip-diskspace-check` -> `--swupd-diskspace-check`

completions/bash/clr-installer
lebensterben: Added.

completions/bash/clr-installer
mdhorn: The installer has a MoM file, but the `/var/lib/swupd/version` referenced in the code for swupd does not exist. We've be better to use `swupd info --quiet` to get the installed version.

This was just a nice-to-have. Do not worry about this one, we can add it later.

completions/bash/clr-installer
mdhorn: For the booleans, the list should be what presented
`--allow-insecure-http|--archive|--cfPurge|--copy-network|--copy-swupd|--high-contrast|--iso|--keep-image|--offline|--reboot|--stub-image|--swupd-clean|--swupd-skip-diskspace-check|--swupd-skip-optional|--telemetry|--tui`

How are you determining the shorter list you are proposing?

I do like the idea of `no`, that is not a part of the standard flag parsing library we are using. Maybe sometime in the future we can upstream a fix.

completions/bash/clr-installer
mdhorn: `--crypt-file` currently does not have any requirements.
 `--log-file` are currently requiring an extension, but .log is the preference.

completions/bash/clr-installer
lebensterben: For the short list, well, I read the source code. I actually learnt Go
language just for writing completions for this last year.

Specifically, I checked `args/args.go`, for all `flag.BoolVar` and
`flag.BoolVarP`.

The followings defaults to true:
- `--reboot`
- `--swupd-skip-diskspace-check`
- `--archive`
- `--keep-image`
- `--copy-network`
And --copy-swupd` defaults to `true` for `interactive` as its description says.

My reasoning is:
- For mass installation with a config file, certainly it's avoids
  confusions to require the user to *explicitly* give the boolean value to
  those options.
- But when the installer is invoked from command line, it's quite rare to
  see patterns like `--foo=true`. Some GNU-style commands has
  `--(en|dis)able-FOO` or `--(no-|)FOO` options, that's a bit verbose but
  avoids any confusions at all.
- IMO, the problem of options accepting boolean arguments and defaults to
  `true` is:
  - Take `--swupd-skip-diskspace-check` for example, it defaults to `true`.
    When users see this, they may expect it to be `opt-in`. But it's
    actually the default behaviour. So they may writing explicitly
    `--swupd-skip-diskspace-check=true` when it's not necessary. Or they
    may not setting `--swupd-skip-diskspace-check=false` because they
    mis-understand how this option works.
  - This is especially an issue for `bash` users, because there's not way
    of showing descriptions in completion menu.
  - Surely users SHOULD read manual first before invoking the installer,
    but making careless mistakes is human nature....
- Therefore, I think that it's only necessary to provide that
  `true or false` choices for those defaults to `true`, to avoid any
  confusion. And for all that defaults to `false`, merely setting the flag
  without an argument is equivalent to setting it to `true`, which is
  concise and easy to understand.

completions/bash/clr-installer
lebensterben: `swupd info | sed '2q;d' | cut -d' ' -f3` can get the version string.
Can you help to verify that `sed` is available in the live ISO (both server and desktop)?

completions/bash/clr-installer
lebensterben: In this case, I will just leave it unchanged.

completions/bash/clr-installer
mdhorn: Okay, only offering the true or false option to booleans that default to true is fine.

completions/bash/clr-installer
mdhorn:
```
swupd search /usr/bin/sed
Bundle with the best search result:
     os-core                        	- Run a minimal Linux userspace.  (64MB) (installed)
```
sed is is os-core, so it will always be present

mdhorn: > `swupd info | sed '2q;d' | cut -d' ' -f3` can get the version string.
> Can you help to verify that `sed` is available in the live ISO (both server and desktop)?

Why add the two pipes, sed, and cut if `swupd info --quiet` does the job?

lebensterben: I just burned a live iso with latest version.

Unfortunately,
- Running `swupd` gives `Input/output error`. So `swupd info --version`
  won't work
- There's no `/var/lib/swupd`.

Can you reproduce this?

mdhorn: Okay, let's drop this from the PR.
The content is there in the ISO I generate when testing the installer; Dev Ops must be using the clean swupd option to reduce the image time.

karthikprabhuvinod:
@lebensterben sorry about the delay..will try this tomorrow

karthikprabhuvinod:
I would like it you name it as `clr-installer.bash` and `_clr-installer.zh` to help find them in future.

As a suggestion, I would like if we can store these in root directory with above names so that it is easy to source them.

mdhorn:
> I would like it you name it as `clr-installer.bash` and `_clr-installer.zh` to help find them in future.
> 
> As a suggestion, I would like if we can store these in root directory with above names so that it is easy to source them.
> 
> @mdhorn What do you think?

These files are not directly sourced by the user, they are installed and read by the shell when launched. They are currently named okay.
See my patch regarding updates to the Makefile.  If there are no other changes, I'll merge this and add the Makefile patch myself.

mdhorn:
@lebensterben Thank you for the PR!

PR #728, opened 2020-05-08
Fix case of allowInsecureHTTP for YAML config

mdhorn:
Documentation and standard convention requires allowInsecureHTTP to be camel-case, but the code and documentation were both incorrect. 
Some users will need to update the YAML files if they already had the incorrect 'AllowInsecureHttp' which is now 'allowInsecureHTTP'.

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -67,7 +67,7 @@ type SystemInstall struct {
 	Kernel            *kernel.Kernel                   `yaml:"kernel,omitempty,flow"`
 	PostReboot        bool                             `yaml:"postReboot,omitempty,flow"`
 	SwupdMirror       string                           `yaml:"swupdMirror,omitempty,flow"`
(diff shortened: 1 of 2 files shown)


mdhorn:
@mbelluzzo @gtkramer Are either of you aware of the use of the allowInsecureHTTP in YAML files used by DevOps?
These will need to be updated.

mbelluzzo:
> @mbelluzzo @gtkramer Are either of you aware of the use of the allowInsecureHTTP in YAML files used by DevOps?
> These will need to be updated.

I couldn't find any on our .yaml files.

mdhorn:
> > @mbelluzzo @gtkramer Are either of you aware of the use of the allowInsecureHTTP in YAML files used by DevOps?
> > These will need to be updated.
> 
> I couldn't find any on our .yaml files.

Good to know. If you are okay with this change, can I get you to approve this PR.
It would have been most likely something seen by customers updating from an HTTP only hosted mix.

mdhorn:

PR #729, opened 2020-05-08
yaml: switch to strict unmarshal

mdhorn:
- Enable Strict Unmarshal to ensure that we fail on unknown tokens and duplicate tokens.


mdhorn:
@mbelluzzo @gtkramer Just FYI, coming in the next installer release.
Already checked that the production DevOps YAMLs are all passing and do not require any changes.

mdhorn:

PR #731, opened 2020-05-09
storage: Fix legacy mode mkfs bug

mdhorn:
- When adding the required options to mkfs for legacy mode on ext, an extra null was being added when options field was empty in the YAML.


mdhorn:

PR #736, opened 2020-05-16
Fix usage model for offline content

mdhorn:
- Always use offline content by default
- Ensure network check is run before install if offline content won't be used

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -114,7 +114,10 @@ func Install(rootDir string, model *model.SystemInstall, options args.Args) erro
 	}
 
 	// Using MassInstaller (non-UI) the network will not have been checked yet
-	if !NetworkPassing && !options.StubImage && !swupd.IsOfflineContent() && len(model.UserBundles) != 0 {
+	if !NetworkPassing &&
+		!options.StubImage &&
+		!swupd.OfflineIsUsable(version, options) &&
+		len(model.UserBundles) != 0 {
 		if err = ConfigureNetwork(model); err != nil {
 			return err
 		}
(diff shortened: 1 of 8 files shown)


swupd/swupd.go
```diff
@@ -701,3 +701,29 @@ func CopyConfigurations(rootDir string) {
...
+	if IsOfflineContent() {
+		if err := utils.ParseOSClearVersion(); err != nil {
+			return usable
+		}
+
+		// If command line did not explicitly set "latest", use the current
+		// image version to enable Offline content usage
+		// Note: Setting "version: 0" in the YAML will not override offline content
+		if utils.IsLatestVersion(version) && options.SwupdVersion == "" {
+			version = utils.ClearVersion
```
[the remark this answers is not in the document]
mdhorn: On line 720? No.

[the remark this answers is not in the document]
karthikprabhuvinod: Well you are using utils.ClearVersion to version and then @723 checking for equality which is going to match and then assign usable to true & return usable...which returns true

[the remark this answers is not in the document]
mdhorn: Yes, technically we can return true there.

I guess I'm old school thinking, where we usually want one true/positive/works return at the end of the function -- and you might short-circuit by returning early if you have a known error or false condition.

We could also remove the `usable` variable and just return true or false everywhere.
What I do not like about that is if you decide to add more login later that is required to meet in order for 'true' to be returned you have to refactor the code and logic flow anywhere you had a return true.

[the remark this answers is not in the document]
mdhorn: @karthikprabhuvinod If this you that's more readable; I'll try it.

swupd/swupd_test.go
```diff
@@ -196,3 +198,44 @@ func TestProcess(t *testing.T) {
...
+
+	if err := utils.ParseOSClearVersion(); err != nil {
+		t.Fatalf("Could not parse OS version")
+	}
+
+	if OfflineIsUsable("latest", options) {
+		t.Fatalf("Offline Content should not be usable to no content dir")
+	}
+
+	if !utils.IsRoot() {
```
karthikprabhuvinod, line 215: Can we move this to top of the test?

swupd/swupd_test.go
```diff
@@ -196,3 +198,44 @@ func TestProcess(t *testing.T) {
...
+		t.Skip("Not running as 'root', skipping test")
+	}
+
+	if err := utils.MkdirAll(conf.OfflineContentDir, 0700); err != nil {
+		t.Fatalf("Could not make offline content dir: %v", err)
+	}
+	defer func() { _ = os.RemoveAll(conf.OfflineContentDir) }()
+
+	if OfflineIsUsable("latest", options) {
+		t.Fatalf("Offline Content should not be usable to due %s", options.SwupdVersion)
```
karthikprabhuvinod, line 225: typo: should be `due to`

swupd/swupd_test.go
```diff
@@ -196,3 +198,44 @@ func TestProcess(t *testing.T) {
...
+	}
+
+	options.SwupdVersion = ""
+	version := "1000"
+	if OfflineIsUsable(version, options) {
+		t.Fatalf("Offline Content should not be usable to due version %s", version)
```
karthikprabhuvinod, line 231: typo: should be `due to`

swupd/swupd_test.go
```diff
@@ -196,3 +198,44 @@ func TestProcess(t *testing.T) {
...
+	options := args.Args{
+		SwupdVersion: "latest",
+	}
+
+	if err := utils.ParseOSClearVersion(); err != nil {
+		t.Fatalf("Could not parse OS version")
+	}
+
+	if OfflineIsUsable("latest", options) {
+		t.Fatalf("Offline Content should not be usable to no content dir")
```
karthikprabhuvinod, line 212: typo: Offline Content should not be usable

controller/controller.go
```diff
@@ -569,27 +569,36 @@ func contentInstall(rootDir string, version string,
...
+		// image version to enable Offline content usage
+		// Note: Setting "version: 0" in the YAML will not override offline content
+		if version == "latest" && options.SwupdVersion == "" {
+			version = utils.ClearVersion
+			log.Info("Overriding version from 'latest' to %s to enable offline install", utils.ClearVersion)
+		}
+
+		if version == utils.ClearVersion {
+			msg := utils.Locale.Get("Copying cached content to target media")
+			prg = progress.NewLoop(msg)
```
karthikprabhuvinod: I know we used to do that before, but just pointing out, we will be reporting progress success for offline content copy even if copyOfflineToStatedir operation has an error(I know its a soft error but it gives ). 

This is just for discussion. No action needed

karthikprabhuvinod:
@mdhorn I tested your PR and it looks good. Some minor feedback and observations

controller/controller.go
mdhorn: Updated to show a failure, but continue.

TUI:
cache_fail_tui

GUI:
cache_fail_gui

swupd/swupd_test.go
mdhorn: It drops the coverage for non-root testing by almost 1%, but moved.

mdhorn: moved back because of the coverage numbers.

Issue #490, opened 2019-08-06
Would be nice to have "Check integrity of installer" option prior to starting it


bktan8:
It would be nice to have the ability to run an integrity check of the installer.  I know you can do a checksum of the image itself, but having ability to check the actual bits that's been burned to a device would be great.  Have this as a menu item at the systemd boot menu.

lebensterben:
>It would be nice to have the ability to run an integrity check of the installer.  I know you can do a checksum of the image itself, but having ability to check the actual bits that's been burned to a device would be great.  Have this as a menu item at the systemd boot menu.  

@bktan8

It seems unnecessary to me to have this integrity check.

A checksum could make sure that the live image you downloaded is not compromised. And when you burn the live image with `dd` command, it's very unlikely that the burned image is different from the is file. Because `dd` is making byte-to-byte copies.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

karthikprabhuvinod:
@lebensterben Good point. However, if there are hiccups during the generation process, it could leave a bad image which is not really bootable. There is no way of knowing that unless we have a source media integrity check.

mdhorn:
Fixed in clr-installer 2.6.1, Clear Linux OS 33300 or higher.

PR #737, opened 2020-05-19
ISO: Add integrity check

karthikprabhuvinod:
Fixes Issue: #490 

- [x] Adding Implant code
- [x] initramFs Add verify checksum using checkisomd5
- [x] BIOS/UEFI menu changes
- [x] Template changes during InitRd making & creating isolinux boot menus
- [x] While making initRd, make swupd install clr-installer-iso-init instead of clr-init
- [x] Adding a new conf file to have a UEFI menu to trigger a ISO media integrity

### DIFF
--- a/args/args.go
+++ b/args/args.go
@@ -33,7 +33,9 @@ const (
 	kernelCmdlineDemo         = "clri.demo"
 	kernelCmdlineLog          = "clri.loglevel"
 	kernelCmdlineHighContrast = "clri.hc"
-	logFileEnvironVar         = "CLR_INSTALLER_LOG_FILE"
+	// KernelMediaCheck is used to create a verufy ISO media boot menu
+	KernelMediaCheck  = "clri.mediacheck"
+	logFileEnvironVar = "CLR_INSTALLER_LOG_FILE"
 )
 
 var (
--- a/iso_templates/initrd_init_template
+++ b/iso_templates/initrd_init_template
@@ -8,14 +8,36 @@ WARNING="\\033[1;33m"        # yellow
 FAILURE="\\033[1;31m"        # red
 INFO="\\033[1;36m"           # light cyan
 
+TTY_DEFAULT="/dev/console"
+TTYS="/dev/console /dev/tty0 /dev/ttyS0 /dev/tty1 /dev/ttyS1 /dev/tty2 /dev/ttyS2"
+TTY="${TTY_DEFAULT}"
+
+discover_tty() {
+    found=0
+    for tty in ${TTYS}
+    do
+        echo -e "" >> "${tty}" 2>/dev/null
+        if [ "$?" -eq 0 ] ; then
+            TTY="${tty}"
+            found=1
+            break
+        fi
+    done
+
+    if [ "${found}" -eq 1 ] ; then
+        echo_tty_kmsg "${INFO}Using ${TTY} for output${NORMAL}"
+    else
+        TTY="${TTY_DEFAULT}"
+        echo_tty_kmsg "${FAILURE}Defaulting to ${TTY} for output${NORMAL}"
+    fi
+}
+
 # Prints msg to TTY and kmsg, stripping color codes for kmsg
 # This echos twice - once for tty0 and once on kmsg for all others
 # $1 - Message string to print
 # Explicitly send boot messages to tty0, /dev/console is ttyS0
 echo_tty_kmsg() {
-    echo -e "$1" > /dev/tty0
-    # This is required for messages to be seen while using qemu
-    echo -e "$1" > /dev/console
+    echo -e "$1" > "${TTY}"
     echo -e "$1" | sed 's/\x1b\[[0-9;]*m//g' > /dev/kmsg
 }
 
@@ -128,12 +150,11 @@ verify_media() {
     local need="ISO integrity check"
 
     if [ -n "${installer}" ]; then
-        checkisomd5 --verbose ${installer} 2>&1 | tee -a /dev/console > /dev/tty0
+        checkisomd5 --verbose ${installer} |& tee -a "${TTY}"
+        check_iso_verify "${PIPESTATUS[0]}" "$need"
     else
         shell_trap "[${FAILURE} FAIL ${NORMAL}] Failed to verify installer media, failed to boot Clear Linux*."
     fi
-
-    check_iso_verify "${PIPESTATUS[0]}" "$need"
 }
 
 # Finds the installer media
@@ -196,6 +217,8 @@ main() {
     mount -t sysfs none /sys
     mount -t devtmpfs none /dev
     mount -t tmpfs none /run
+
+    discover_tty
 
     # Verify CPU features needed to run Clear exist
     echo_tty_kmsg "Checking if system is capable of running Clear Linux*..."
--- a/iso_templates/initrd_init_template
+++ b/iso_templates/initrd_init_template
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Set color commands used with echo
 # Refer to man console_codes for more info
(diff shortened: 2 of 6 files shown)


karthikprabhuvinod:
@mdhorn Hey Mark, this is just the first round to get your feedback. I flashed this a USB with the image and I ran into a Wront EFI laoder issue similar to this one:

https://community.clearlinux.org/t/usb-install-error-on-xps-13-9370-wrong-efi-loader-signature/3100

I am going to do some more tests.

karthikprabhuvinod:
@mdhorn what command should we run when the user provides a keystroke , reboot and # - systemctl reboot both hang the init session😕

mdhorn:
> @mdhorn what command should we run when the user provides a keystroke , reboot and # - systemctl reboot both hang the init session😕

Then for now, just implement what we know works.
- For Failure, use the 'shell_trap'
- For Success, just have 'Press any key to boot into live OS'; and just let the boot process continue

If we have high interest later, we can investigate the reboot option.

karthikprabhuvinod:
@mdhorn Looks like 33180 has been released. So I will change clr-core-plus to clr-installer-os-init

args/args.go
```diff
@@ -33,6 +33,7 @@ const (
 	kernelCmdlineDemo         = "clri.demo"
 	kernelCmdlineLog          = "clri.loglevel"
 	kernelCmdlineHighContrast = "clri.hc"
+	kernelMediaCheck          = "clri.mediacheck"
```
mdhorn, line 36: Since this value is needed in the iso utils code, perhaps it should be exported and used in isoutils/isoutils.go for consistency. 
```suggestion
	KernelMediaCheck          = "clri.mediacheck"
```

args/args.go
```diff
@@ -121,6 +123,8 @@ func (args *Args) setKernelArgs() (err error) {
 			args.DemoMode = true
 		} else if strings.HasPrefix(curr, kernelCmdlineHighContrast) {
 			args.HighContrast = true
+		} else if strings.HasPrefix(curr, kernelMediaCheck) {
```
mdhorn, line 126:
```suggestion
		} else if strings.HasPrefix(curr, KernelMediaCheck) {
```

isoutils/isoutils.go
```diff
@@ -320,6 +320,21 @@ func buildInitrdImage() error {
...
+// We take a copy of options from normal EFI entry and edit it for new EFI entry: /loader/entries/iso-checksum.conf
+func prepareISOCheckSumBootEntry(inputOptions []string) []string {
+	outputOptions := append([]string(nil), inputOptions...)
+	for i, option := range outputOptions {
+		if strings.Contains(option, "title") {
+			outputOptions[i] = "title Verify ISO Integrity"
+		}
+
+		if i == len(outputOptions)-1 {
+			outputOptions[i] = strings.TrimSpace(option) + " clri.checkmedia"
```
mdhorn, line 332:
```suggestion
			outputOptions[i] = strings.TrimSpace(option) + " " + args.KernelMediaCheck
```

iso_templates/initrd_init_template
```diff
@@ -104,6 +104,31 @@ mount_root() {
...
+        if [ "$?" = 0 ] ; then
+            break
+        fi
+    done
+}
+
+check_iso_verify() {
+    local ret="$1"
+    local msg="$2"
+    [ "$ret" -ne 0 ] && { echo_tty_kmsg "[${FAILURE} FAIL ${NORMAL}] $msg: Press any key to reboot your system"; wait_for_keystrokes; shell_trap "$msg"; }
```
mdhorn, line 119: We don't want Press a key message on failures, only for success before continuing the boot.

```suggestion
    [ "$ret" -ne 0 ] && { echo_tty_kmsg "[${FAILURE} FAIL ${NORMAL}] $msg; shell_trap "$msg"; }
```

iso_templates/initrd_init_template
```diff
@@ -104,6 +104,31 @@ mount_root() {
...
+    [ "$ret" -eq 0 ] && { echo_tty_kmsg "[${SUCCESS}  OK  ${NORMAL}] $msg"; }
```
mdhorn, line 120:
```suggestion
    [ "$ret" -eq 0 ] && { echo_tty_kmsg "[${SUCCESS}  OK  ${NORMAL}] $msg: Press any key to boot to the Live OS"; wait_for_keystrokes;"; }
```

karthikprabhuvinod: Done

iso_templates/initrd_init_template
karthikprabhuvinod: Done

isoutils/isoutils.go
karthikprabhuvinod: Done

args/args.go
karthikprabhuvinod: Done

args/args.go
karthikprabhuvinod: Done

karthikprabhuvinod:
@mdhorn: did the refactoring finding and mounting installer. Also a bit of cleanup. Still not able to boot on Bare metal. 

does your command have some magic flags i am not aware of

i use plain old : sudo dd if=~/path/toiso of=/dev/sdb status=progress

/dev/sdb has not even a partition table

karthikprabhuvinod:
@mdhorn : I was able to get the messages being print out both in cases of qemu and bare metal.

About the Press any key to continue feature. Looks like the keyboard is not ready at the time of loading initramfs. However it can be done, by looking at the post below. Apparently there is a /etc/mkinitcpio.conf available where the keyboard to be added as a HOOK.

[Reference link](https://wiki.archlinux.org/index.php/Mkinitcpio)

**Excerpt from the link:**

**PS/2 keyboard users:** In order to get keyboard input during early init, if you do not have it already, add the keyboard hook to the HOOKS. On some motherboards (mostly ancient ones, but also a few new ones), the i8042 controller cannot be automatically detected. It is rare, but some people will surely be without keyboard. You can detect this situation in advance. If you have a PS/2 port and get i8042: PNP: No PS/2 controller found. Probing ports directly message, add atkbd to the MODULES.

mdhorn:
> @mdhorn : I was able to get the messages being print out both in cases of qemu and bare metal.
> 
> About the Press any key to continue feature. Looks like the keyboard is not ready at the time of loading initramfs. However it can be done, by looking at the post below. Apparently there is a /etc/mkinitcpio.conf available where the keyboard to be added as a HOOK.
> 


We should already have a bundle to support keyboards.
Trying making a mix, and add `kbd` to the bundle list for `clr-installer-iso-init`.

karthikprabhuvinod:
> > @mdhorn : I was able to get the messages being print out both in cases of qemu and bare metal.
> > About the Press any key to continue feature. Looks like the keyboard is not ready at the time of loading initramfs. However it can be done, by looking at the post below. Apparently there is a /etc/mkinitcpio.conf available where the keyboard to be added as a HOOK.
> 
> We should already have a bundle to support keyboards.
> Trying making a mix, and add `kbd` to the bundle list for `clr-installer-iso-init`.

I think Might not even have to do that. I might have to just load the keyboard kernel module during init.

iso_templates/initrd_init_template
```diff
@@ -104,15 +106,45 @@ mount_root() {
     fi
 }
 
-# Finds the installer media and calls mount_root if successful
-find_and_mount_installer() {
+wait_for_keystrokes() {
+    while [ true ] ; do
+        read -t 2 -n 1
+        if [ "$?" = 0 ] ; then
```
mdhorn, line 112:
```suggestion
        if [ "$?" -eq 0 ] ; then
```

mdhorn: instead of the loop forever and the read with timeout, why not just the read?
`read -n 1`

karthikprabhuvinod: I guess I can do that. Also will help saves cpu cycles as it would block on input

mdhorn: > I guess I can do that. Also will help saves cpu cycles as it would block on input

assuming we keep the wait for the input section

karthikprabhuvinod: Done

karthikprabhuvinod:
@mdhorn I tried but couldn't get the keyboard input to work. As we decided, I am going to wait 10secs after a successful integrity check before booting into the root of the live image

iso_templates/initrd_init_template
```diff
@@ -104,15 +128,46 @@ mount_root() {
     fi
 }
 
-# Finds the installer media and calls mount_root if successful
-find_and_mount_installer() {
+wait_for_timeout() {
+    local n="$1"
+    echo_tty_kmsg > "${TTY}"
```
mdhorn, line 133: The redirect is extraneous 
```suggestion
    echo_tty_kmsg
```

iso_templates/initrd_init_template
```diff
@@ -104,15 +128,46 @@ mount_root() {
...
+    echo_tty_kmsg > "${TTY}"
```
mdhorn, line 134: The redirect is extraneous 
```suggestion
    echo_tty_kmsg
```

karthikprabhuvinod: Thanks ..Done

iso_templates/initrd_init_template
karthikprabhuvinod: Thanks ..Done

iso_templates/initrd_init_template
```diff
@@ -104,15 +128,46 @@ mount_root() {
...
+    [ "$ret" -eq 0 ] && { echo_tty_kmsg "[${SUCCESS}  OK  ${NORMAL}] $msg"; wait_for_timeout 10; }
+}
+
+# Check ISO media integrity
+verify_media() {
+    local installer="$1"
+    local need="ISO integrity check"
+
+    if [ -n "${installer}" ]; then
+        checkisomd5 --verbose ${installer} |& tee -a "${TTY}"
```
mdhorn, line 155: Sorry, one other fix
```suggestion
        checkisomd5 --verbose ${installer} > "${TTY}"
```
QEMU was getting the output twice before this change.

karthikprabhuvinod: Yes , I missed it. Will fix , thanks. I saw something to fix too. Will push them together

karthikprabhuvinod: Done

PR #738, opened 2020-05-20
Fix JSON decoding of ister Version

mdhorn:
- Fix how JSON Version field is decoded when reading ister configuration files
- Go 1.14.x now fails json.Number with invalid values

### DIFF
--- a/model/model_ister.go
+++ b/model/model_ister.go
@@ -5,8 +5,10 @@
 package model
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -22,7 +24,12 @@ import (
 	"github.com/clearlinux/clr-installer/storage"
 	"github.com/clearlinux/clr-installer/telemetry"
 	"github.com/clearlinux/clr-installer/user"
+	"github.com/clearlinux/clr-installer/utils"
 )
+
+type UintString struct {
+	Number uint
+}
 
 // IsterConfig represents the install configuration used in the "ister" app
 type IsterConfig struct {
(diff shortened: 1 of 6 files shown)


model/model.go
```diff
@@ -120,6 +120,24 @@ type StorageAlias struct {
 	DeviceFile bool   `yaml:"devicefile,omitempty,flow"`
 }
 
+// InitializeDefaults ensure defaults are set such
+// as initializing pointer members
+func (si *SystemInstall) InitializeDefaults() {
```
karthikprabhuvinod, line 125: We need a rebase correct? I thought you added this as this based on previous PR

mdhorn: This is from the previous commit. 

Only the three new commits are from this PR, but because I pushed the branch before the merge of the previous PR, it looks like 5 commits.

mdhorn: There, I just force pushed and now you should only see the three new commits.

model/model_ister.go
```diff
@@ -366,3 +367,46 @@ func setStorageValues(name string, part uint64, size string) (storage.BlockDevic
...
+	for {
+		token, err := dec.Token()
+		if err == io.EOF {
+			break
+		}
+
+		switch t := token.(type) {
+		case json.Number:
+			var n int64
+
```
karthikprabhuvinod, line 385: May can get rid of this extra space

model/model_ister.go
```diff
@@ -366,3 +367,46 @@ func setStorageValues(name string, part uint64, size string) (storage.BlockDevic
...
+			n, err = t.Int64()
+			if err != nil {
+				us.Number = 0
+				return err
+			}
+
```
karthikprabhuvinod, line 391: here too

model/model_ister.go
mdhorn: updated

model/model_ister.go
mdhorn: update

PR #744, opened 2020-06-08
controller: Fix hang bug for cbm fails

mdhorn:
- Fix installer hang when clr-boot-manager fails

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -662,7 +662,6 @@ func contentInstall(rootDir string, version string,
 		if err := sw.DisableUpdate(); err != nil {
 			warnMsg := utils.Locale.Get("Disabling automatic updates failed")
 			log.Warning(warnMsg)
-			prg.Failure()
 			return prg, err
 		}
 		prg.Success()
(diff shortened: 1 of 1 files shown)


karthikprabhuvinod:
But if CBM fails, dont we need to report a failure too?

mdhorn:
> But if CBM fails, dont we need to report a failure too?

We do, if you follow the return we log and fail there.

mdhorn:
> > But if CBM fails, dont we need to report a failure too?
> 
> We do, if you follow the return we log and fail there.

Or the failure and output from the cbm is logged with
`cmd.RunAndLogWithEnv
`

karthikprabhuvinod:
> > > But if CBM fails, dont we need to report a failure too?
> > 
> > 
> > We do, if you follow the return we log and fail there.
> 
> Or the failure and output from the cbm is logged with
> `cmd.RunAndLogWithEnv `

Okay, but thats in the log file, Do we see a fail on the progress?

mdhorn:
> > > > But if CBM fails, dont we need to report a failure too?
> > > 
> > > 
> > > We do, if you follow the return we log and fail there.
> > 
> > 
> > Or the failure and output from the cbm is logged with
> > `cmd.RunAndLogWithEnv `
> 
> Okay, but thats in the log file, Do we see a fail on the progress?

Yes
https://github.com/clearlinux/clr-installer/blob/master/controller/controller.go#L409

Issue #8, opened 2018-08-13
Support LVM


mdhorn:
<a href="https://github.com/mdhorn"><img src="https://avatars.githubusercontent.com/mdhorn" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mdhorn](https://github.com/mdhorn)**
_Friday Mar 02, 2018 at 22:18 GMT_

----

Future Enhancements: LVM, Resizing partitions

mdhorn:
Note: We will need to add support for LVM to the special initd similar to what was done to support Encryption for the root file systems:

**initrd=\EFI\org.clearlinux\initrd-org.clearlinux.native.4.20.8-696 initrd=\EFI\org.clearlinux\freestanding-clr-init.cpio.gz** root=UUID=919a625a-8e45-4b08-9e97-e4e6bf4621cb **rd.luks.uuid=6cd79a0f-ed3e-47c5-9e81-016e9683e2fe** quiet console=tty0 console=ttyS0,115200n8 init=/usr/bin/initra-desktop  initcall_debug tsc=reliable no_timer_check noreplace-smp kvm-intel.nested=1 rootfstype=ext4,btrfs,xfs intel_iommu=igfx_off cryptomgr.notests rcupdate.rcu_expedited=1 rcu_nocbs=0-64 rw **rootflags=x-systemd.device-timeout=0**

NamPNQ:
Can you tell me how to use lvm?

mdhorn:
When LVM support is enabled for the Installer, we need to ensure we can overwrite/reuse existing disks which already have LVM enabled.

Currently, if we attempt to overwrite a disk enabled with LVM, it will fail:
Steps to Repro:

1. Install CentOS and choose automatic partition options
2. Start Clear installer
3. Choose "Destructive Installation"
4. Proceed to install and it will fail

 To clear the LVMs prior to installing, these steps seem to work:
```
$ sudo sgdisk -Z /dev/<device>
$ sudo partprobe
$ sudo dmsetup remove_all --force
$ sudo partprobe

```

mdhorn:
Clear Linux OS requirement to enable LVM
https://github.com/clearlinux/distribution/issues/912

mdhorn:
Ensure we have software RAID support also while implementing LVM support.

mdhorn:
Blocked by #530

mdhorn:
[...]

PR #746, opened 2020-06-11
Enable lvm root

mdhorn:
Fixes Issue: #8 

- Enable the use of LVM for root partition during installation
- Allow additional Advanced Installation partitions to be logical volumes (UNDER DEBUG)

### DIFF
--- a/storage/block_devices.go
+++ b/storage/block_devices.go
@@ -23,6 +23,7 @@ import (
 type BlockDevice struct {
 	Name            string             // device name
 	MappedName      string             // mapped device name
+	Path            string             // device path
 	Model           string             // device model
 	MajorMinor      string             // major:minor device number
 	PtType          string             // partition table type
@@ -263,6 +264,10 @@ func (bd *BlockDevice) GetPartitionNumber() uint64 {
 
 // GetDeviceFile formats the block device's file path
 func (bd BlockDevice) GetDeviceFile() string {
+	if bd.Path != "" {
+		return bd.Path
+	}
+
 	return filepath.Join("/dev/", bd.Name)
 }
 
@@ -277,7 +282,7 @@ func (bd BlockDevice) GetMappedDeviceFile() string {
 		return filepath.Join("/dev/mapper", bd.Name)
 	}
 
-	return filepath.Join("/dev/", bd.Name)
+	return bd.GetDeviceFile()
 }
 
 // GetDeviceID returns an identifier for the block device
@@ -343,6 +348,7 @@ func (bd *BlockDevice) Clone() *BlockDevice {
 	clone := &BlockDevice{
 		Name:            bd.Name,
 		MappedName:      bd.MappedName,
+		Path:            bd.Path,
 		Model:           bd.Model,
 		MajorMinor:      bd.MajorMinor,
 		FsType:          bd.FsType,
--- a/storage/block_devices.go
+++ b/storage/block_devices.go
@@ -110,6 +110,9 @@ const (
 	// BlockDeviceStateConnected identifies a BlockDevice as Connected
 	BlockDeviceStateConnected
 
+	// BlockDeviceStateSuspended identifies a BlockDevice as Suspended
+	BlockDeviceStateSuspended
+
 	// MinimumPartitionSize is smallest size for any partition
 	MinimumPartitionSize = 1048576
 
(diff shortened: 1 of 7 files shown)


mdhorn:
@karthikprabhuvinod Still debugging why the boot hangs on mounting additional logical volumes, but if you have the bandwidth, please start reviewing this changeset.

storage/block_devices.go
```diff
@@ -214,6 +221,19 @@ func getAliasSuffix(file string) string {
 	return ""
 }
 
+func (bd *BlockDevice) FindAllChildren() []*BlockDevice {
```
karthikprabhuvinod, line 224: Instead of this use FindAllBlockDevices from block_devices_filters.go

storage/block_devices.go
```diff
@@ -263,6 +283,10 @@ func (bd *BlockDevice) GetPartitionNumber() uint64 {
 
 // GetDeviceFile formats the block device's file path
 func (bd BlockDevice) GetDeviceFile() string {
+	if bd.Path != "" {
+		return bd.Path
```
karthikprabhuvinod, line 287: Could you make sure that this change does not break make functional-check

storage/block_devices.go
karthikprabhuvinod: I did look at this. I feel the locations where you call this , you have a lot of filter logic and that would become part of inner function in my approach. I tried but I wasn't able to create a much simplified solution with benefits. So I think you can overlook this. The commit i pushed addresses another location where simplification had benefits.

storage/block_devices.go
karthikprabhuvinod: Thanks for confirming this

storage/block_devices.go
mdhorn: @karthikprabhuvinod When testing your commit, it was working fine for the LVM, but it removed media for all Safe and Destructive installations. I've since dropped that commit.

storage/block_devices.go
```diff
@@ -632,6 +660,18 @@ func isBlockDeviceAvailableDeep(blockDevices []*BlockDevice) bool {
 			break
 		}
 
+		// Hack for Logical Volumes
+		// Since lvms do not have Partition Labels, if the Logical Volume name
+		// which matches Advance Labels, force set the Partition Label
+		if bd.Type == BlockDeviceTypeLVM2Volume {
+			log.Debug("isBlockDeviceAvailableDeep: Found lvm name: %s", bd.Name)
+			if matches := lvmAdvancedLabelExp.FindStringSubmatch(bd.Name); len(matches) == 2 {
```
karthikprabhuvinod, line 668: Can we LOG an error to disk if we the regex match failed

mdhorn: Done:
```
2020/06/23 22:52:17 [DBG] isBlockDeviceAvailableDeep: Found lvm name: VolGrpClearRoot-blah
2020/06/23 22:52:17 [DBG] lvm name VolGrpClearRoot-blah does not match Advanced regex \w+-(CLR_[0-9a-zA-Z-_\+.]+$)

```

karthikprabhuvinod:
> Changewise LGTM to mark. I was thinking if you can add a functional test for this. but I will leave it upto you.
> 
> **Functional test:** To add a test we create a LVMs with our desired labels and run advanced partitioning on it and check installation succeededs and fstab has required entries

@mdhorn  any comments?

mdhorn:
> > Changewise LGTM to mark. I was thinking if you can add a functional test for this. but I will leave it upto you.
> > **Functional test:** To add a test we create a LVMs with our desired labels and run advanced partitioning on it and check installation succeededs and fstab has required entries
> 
> @mdhorn any comments?

Functional test pushed.

kinshakov:
Sorry, but I haven't got is it possible to install root on LVM now or not? Because installer doesn't see my lvm partitions. He sees the hole physical drive.

mdhorn:
> Sorry, but I haven't got is it possible to install root on LVM now or not? Because installer doesn't see my lvm partitions. He sees the hole physical drive.

You need to use the Advanced Installation selection for the media and ensure you create Logical Volume Names which match the required labeling syntax:
https://github.com/clearlinux/clr-installer/blob/master/scripts/InstallerYAMLSyntax.md#advanced-installation-media-targets
If you have Logical Volume names which match `CLR_ROOT` and a standard partition label for `CLR_BOOT` (boot can not be on an LVM) the installation will proceed. All other partitional are optional.

PR #781, opened 2022-09-07
Let systemd check kernel command line options

bwarden:
Cleanup for `clr-installer-provision.service`. Trying to avoid extra embedded bash calls -- use systemd's internal capabilties instead of relying on bashisms.

Fixes Issue: #

- Avoid using bash calls as much as possible
- Use systemd's built-in ConditionKernelCommandLine directive to detect clri.descriptor to decide whether to run
- Allow this unit to take over the console (was blocking, attempting askpass, while getty was active)
- Call cat directly to print /etc/issue to console (instead of from bash)
- Call clr-installer directly instead of from a bash loop -- should implement retries via systemd's mechanisms instead
- Call reboot directly as a post action only if/after the installer succeeded

### DIFF
--- a/etc/systemd/clr-installer-provision.service
+++ b/etc/systemd/clr-installer-provision.service
@@ -20,7 +20,8 @@ ConditionKernelCommandLine=clri.descriptor
 [Service]
 Type=oneshot
 ExecStartPre=-/usr/bin/cat /etc/issue
-ExecStart=/bin/bash -l -c 'until /usr/bin/clr-installer ; do echo "clr-installer failed...retying"; sleep 5 ; done ; /usr/bin/reboot'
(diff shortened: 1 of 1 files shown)


bryteise:
Seems fairly reasonable to me.

ahkok:
It looks fine and a lot cleaner. I'm not sure if we want to `Restart=on-failure` since the default value for that is `no` and it will reboot the system immediately? I would definitely experiment with making clr-installer fail and see whether at a minimum there is some usable output before the system reboots.

But aside that this is a good cleanup.

bwarden:
Also, the console failures I was seeing are possibly because I was on ttyS1, which isn't listed in Conflicts. Worth further testing in a controlled environment.

PR #790, opened 2023-02-22
Show found partitions while we're looking for CLR_ISO

bwarden:
initrd calls blkid a few times looking for CLR_ISO, so it can identify the root device that was booted. On each retry, call blkid plain to show everything that's been found -- this may help with debugging boot failures.

### DIFF
--- a/iso_templates/initrd_init_template
+++ b/iso_templates/initrd_init_template
@@ -177,6 +177,7 @@ find_installer() {
             sleep 1
             (( retries++ ))
         fi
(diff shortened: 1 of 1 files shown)


fenrus75 (approved):

bryteise:
Should this use the `echo_tty_kmsg` to write to both?

bwarden:
> Should this use the `echo_tty_kmsg` to write to both?

Sure, this should do it.

